PR #60, opened 2026-04-23
Proposal: .env.local.example + preflight drift check (for discussion)

### LINKED ISSUE #70: Prevent silent .env.local drift when new required env vars are added
## Problem
`.env.local` is gitignored, so adding a new required env var silently rots old dev setups. We hit this with the e2e auth vars (`CI_RESET_TOKEN`, `E2E_*_PASSWORD`, `SUPABASE_SECRET_KEY`) — old `.env.local` files kept working for `npm run dev` but failed ten layers deep in Playwright setup, with errors that didn't point at the real cause.

## Desired outcome
A committed canonical key list plus a preflight check that catches drift early and points devs at a non-destructive fix (append missing keys) rather than a destructive one (regenerate). CI should exercise the same preflight so the story is uniform.

## Tracking
Work in progress in #60 (env-example-preflight draft).

NorsemanSpiff:
Closes #70

## Context
Opened as a draft for team discussion — happy to close if folks prefer a different approach or no change at all.

## Problem
`.env.local` is gitignored, so adding a new required env var silently rots old dev setups. We hit this with the e2e auth vars (`CI_RESET_TOKEN`, `E2E_*_PASSWORD`, `SUPABASE_SECRET_KEY`) — old `.env.local` files kept working for `npm run dev` but failed ten layers deep in Playwright setup, with errors that didn't point at the real cause.

## Proposal
- Committed `.env.local.example` as the canonical key list.
- `npm run setup` → `copyFileSync` (no longer holds the template inline).
- New `scripts/check-env.mjs` runs first in `npm run dev:db`, diffs keys, fails with an actionable error that prefers the non-destructive fix (append missing keys) over the destructive one (regenerate).
- `ci.yml` gains one line: `cp .env.local.example .env.local` before the functional test run, so preflight passes there.
- Extras in `.env.local` are preserved throughout.

See `docs/devjournal.md` for the full write-up.

## Alternatives considered
- **Merge-append in `setup.mjs`** — rejected because devs don't re-run setup after onboarding; writing to `.env.local` on every `dev:db` is a gross side-effect.
- **Tolerant check (file OR `process.env`)** — rejected because CI would still need to declare all 7 keys somewhere, giving up the simplicity win.
- **Skip check-env in CI** — rejected because it hides behavior from anyone reading `ci.yml`; the explicit `cp` line tells the whole story.
- **`dotenv-safe` / zod runtime schema** — fine patterns, rejected as dep/scope creep for a 60-line problem.

## Test Plan
(See the commit message — fresh onboarding, drift detection + both fix paths, missing-file path, extras preservation, lint, full `npm test`. CI functional job is the one thing not validatable locally — that's what this PR's checks will confirm.)

## Questions for the team
- **Is this pain worth the infrastructure?** It's hit us once so far. Do we expect new env vars often enough to justify the scaffolding, or is "ping the team on WhatsApp + add a devjournal entry when adding one" enough?
- Is the added `ci.yml` step acceptable, or would you prefer a different CI strategy?
- Descriptive-placeholder convention for new keys — reasonable?
- Any objections to the extra preflight step slowing down `dev:db` by ~20ms?


james-baker (approved):
I like it!

Issue #41, opened 2026-04-20
Supabase RLS policies audit


oolu4236:
**Problem**: Without correct RLS policies on public.profiles and the invites table, any authenticated user can query any other user's data via the Supabase client. 

**Goal**: Audit and implement row-level security policies for profiles, invites, and programs. add `authenticated can read their own row, nothing else` Without this, users could access each other’s data. 

```
-- Allow users to read and update only their own profile row
ALTER TABLE public.profiles ENABLE ROW LEVEL SECURITY;

CREATE POLICY "users can read own profile"
ON public.profiles FOR SELECT
USING (auth.uid() = id);

CREATE POLICY "users can update own profile"
ON public.profiles FOR UPDATE
USING (auth.uid() = id);

-- Invites: members can read invites they created or redeemed
CREATE POLICY "members can read own invites"
ON public.invites FOR SELECT
USING (auth.uid() = created_by OR auth.uid() = redeemed_by);
```

james-baker:
I'm not intending to implement RLS - data authorization will be handled in the backend code underneath the Hono API surface.

Hmm... can logged-in users use the Supabase client directly to access the database? I assumed no, but this seems worth checking, because we don't mean to give everyone direct DB access.

james-baker:
"any authenticated user can query any other user's data via the Supabase client" was already not true due to being disabled in the Supabase UI, but now there is "RLS that denys everything" on all the tables as defense-in-depth. Implemented in #76 and documented in #74.

PR #76, opened 2026-04-25
Enable RLS on all public tables (defense-in-depth)

james-baker:
## Summary

- Adds `.enableRLS()` to all four `public` tables in `src/server/schema.ts` (`profiles`, `programs`, `profile_programs`, `invites`).
- Generated migration `0002_clever_radioactive_man.sql` runs `ALTER TABLE ... ENABLE ROW LEVEL SECURITY` on each.
- No policies are added: with RLS on and zero policies, Postgres denies the `anon` and `authenticated` roles by default. The Hono API connects as `postgres` (superuser), which bypasses RLS, so the app is unaffected.

This is a backstop. The hosted Supabase Data API is already disabled at the project level (see `docs/doc-supabase.md`), so neither PostgREST nor pg_graphql can reach these tables today. RLS makes sure that if the Data API toggle ever flips on, the publishable key still can't read or write anything.

The architecture and Supabase docs already describe this state — this PR brings the schema in line with the docs.

## Test plan

- [x] `npm test` (45 functional + 16 e2e) green locally
- [x] Migration applied cleanly to local DB
- [ ] CI green on PR
- [ ] Vercel preview deploy succeeds (preview shares the prod DB; the migration won't run there, but no app code changed)

### DIFF
--- a/drizzle/0002_clever_radioactive_man.sql
+++ b/drizzle/0002_clever_radioactive_man.sql
@@ -1,0 +1,4 @@
+ALTER TABLE "invites" ENABLE ROW LEVEL SECURITY;--> statement-breakpoint
+ALTER TABLE "profile_programs" ENABLE ROW LEVEL SECURITY;--> statement-breakpoint
+ALTER TABLE "profiles" ENABLE ROW LEVEL SECURITY;--> statement-breakpoint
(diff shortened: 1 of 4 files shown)


james-baker:
Oh I forgot to link issue #41 here to see if it automatically closed.

Issue #65, opened 2026-04-23
Add SECURITY.md


oolu4236:
**Problem:**
The repo is public. Anyone who finds a vulnerability has no clear way to report it privately — their only option is opening a public issue, which exposes the problem to everyone before it's fixed. There's also no stated expectation on our end for how we'll respond or how long it takes.

**Goals:**

- Add a SECURITY.md at the repo root with a responsible disclosure policy: where to report, what to include, what to expect from us
- Enable GitHub's private vulnerability reporting (Security tab → Private vulnerability reporting) so reporters have a confidential channel

james-baker:
This is item number 5 on the security hardening plan of #58 . (Leaving here, it's fine to be its own issue addressed separately.)

james-baker:
PR #85 seems to have replaced #80

james-baker:
Taking care of this, changing as described in my review of #85

Issue #58, opened 2026-04-22
Security hardening: public-repo TODO


james-baker:
Remaining TODO items from [`docs/plan-security-hardening.md`](docs/plan-security-hardening.md). Items 3 (hono bump) and 4 (Dependabot + CodeQL) are already merged.

- [x] **Sentry PII over-collection (item 1)** — drop `sendDefaultPii` in `instrumentation-client.ts` and `sentry.server.config.ts`; drop `includeLocalVariables`; pass `replayIntegration({ maskAllText: true, blockAllMedia: true })`. Currently Sentry captures request bodies with `bio`/`liveDesire`/`emergencyContact`/`supplementaryInfo`, full headers + cookies, Replay DOM with unmasked textareas, and local server variables.
- [x] **Sentry Replay on `/auth/callback` (item 2)** — disable Replay on the callback route so PKCE `?code=` URLs and the localStorage verifier aren't shipped to a third party on error.
- [x] **`SECURITY.md` + private vulnerability reporting (item 5)** — add `SECURITY.md` and enable Settings → Security → Private vulnerability reporting before first outside contact.

- [ ] **`E2E_ADMIN_PASSWORD` controls a real prod admin (item 6)** — either drop the long-lived `e2e-admin` account (flip the flag in seed instead) or restrict the e2e workflow with `permissions: read-all` + CODEOWNERS gate on workflow changes.
- [x] **Security headers (item 8)** — add CSP, HSTS, `X-Frame-Options`, `Referrer-Policy` via Next.js `headers()` config or `vercel.json`. CSP is the high-value one.
- [ ] **Length-limit profile fields (item 9)** — `MAX_BIO=10000`, `MAX_DISPLAY_NAME=100`, etc. in `parseEditableProfile`.
- [x] **`/logout` GET → POST (item 10)** — replace link with `<form action="/logout" method="post">` to close the CSRF surface.
- [ ] **Document `sb_secret_…` tokens in git history (item 11)** — README note + `# pragma: allowlist secret`-style annotation so GitGuardian / TruffleHog don't open issues on day one public.

Full rationale and context in `docs/plan-security-hardening.md`.

PR #80, opened 2026-04-26
Add SECURITY.md with private vulnerability reporting policy

oolu4236:
**Problem:**
The repo is public with no disclosure path. Anyone finding a vulnerability has no choice but to open a public issue.

**Goals:**

Committed `SECURITY.md` with where to report, what to include, and what to expect from us
GitHub will surface the "Report a vulnerability" button on the Security tab once private vulnerability reporting is enabled in repo settings (Security → enable "Private vulnerability reporting" — one toggle, with no code change needed)

Ref: Closes part of #58 (item 3) and #65

### DIFF
--- a/scripts/seed-dev.ts
+++ b/scripts/seed-dev.ts
@@ -84,7 +84,8 @@ async function seedAuthUsers(): Promise<SeedResult> {
         id, instance_id, aud, role, email,
         raw_app_meta_data, raw_user_meta_data,
         email_confirmed_at, created_at, updated_at,
-        is_sso_user, is_anonymous
+        is_sso_user, is_anonymous,
+        confirmation_token, recovery_token, email_change_token_new, email_change
       ) VALUES (
         ${profile.id}::uuid,
         '00000000-0000-0000-0000-000000000000'::uuid,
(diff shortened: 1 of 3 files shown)


scripts/seed-dev.ts
```diff
@@ -84,7 +84,8 @@ async function seedAuthUsers(): Promise<SeedResult> {
         id, instance_id, aud, role, email,
         raw_app_meta_data, raw_user_meta_data,
         email_confirmed_at, created_at, updated_at,
-        is_sso_user, is_anonymous
+        is_sso_user, is_anonymous,
```
james-baker, line 87: Why are there seed-dev changes in this PR? I think this PR is superseded by #85 but this one has other non-md changes, so flagging to your attention @oolu4236 to double-check on how your intent ended up here.

james-baker (changes requested):

scripts/seed-dev.ts
oolu4236: @james-baker   Looks like #85 has a cleaner version of SECURITY.md (rotation guide, admin account risk section). and the seed fix is already in main via #79.  The seed-dev.ts change in this PR is the GoTrue token-column fix that was already merged separately via #79. This PR is okay to close

PR #82, opened 2026-04-26
Change /logout from GET to POST to prevent CSRF sign-out

oolu4236:
Cross-site requests (img tags, link prefetch) can trigger GET routes; POST requires an explicit form submission which browsers don't make cross-origin. The home page sign-out button already posts via a Server Action — this aligns the route handler with that model and removes the GET surface. E2e test updated to drive the button instead of navigating directly to /logout.


james-baker:
Looks good to me

PR #84, opened 2026-04-26
Remove sendDefaultPii and tighten Sentry replay/header capture

oolu4236:
## Summary
- Removes `sendDefaultPii: true` from both client and server Sentry configs (was auto-sending cookies, auth headers, user IPs)
- Server `beforeSend` now explicitly strips `cookie` and `authorization` headers from all events
- Client replay configured with `maskAllText: true` + `blockAllMedia: true`
- Client `beforeSend` drops events from `/auth/`, `/login`, `/signup` routes where tokens may appear in the URL

## Test plan
- [x] Functional tests pass
- [x] Trigger a test error in the app and verify no cookies/auth headers appear in Sentry event payload
- [ ] Verify replay recordings mask input text

### DIFF
--- a/docs/doc-sentry.md
+++ b/docs/doc-sentry.md
@@ -26,6 +26,19 @@ Auth tokens are generated at Settings → Auth Tokens in the Sentry dashboard.
 - **Source maps:** uploaded automatically during `next build` when `SENTRY_AUTH_TOKEN` is set
 - **Tunnel route:** `/monitoring` proxies Sentry events through our domain to avoid ad blockers
 
+## PII Handling
+
+`sendDefaultPii` is **off** on both client and server. With it enabled, Sentry's Next.js SDK auto-attaches request cookies, the `Authorization` header, and the user's IP to every event — all of which expose session tokens for an authenticated app.
+
+Two `beforeSend` scrubbers in `src/lib/sentry-scrub.ts` provide defense-in-depth:
+
+- **Client** (`scrubClientEvent`): on `/auth/`, `/login`, and `/signup` URLs, the query string is dropped from `event.request.url` so OAuth tokens like `?code=...` don't ship. The event itself is still sent so genuine errors on these routes remain visible.
+- **Server** (`scrubServerEvent`): deletes `event.request.cookies` and removes the `authorization` and `cookie` headers from every event.
+
+Session replay is configured with `maskAllText: true` and `blockAllMedia: true`, so recordings show only structural placeholders for text and no images/video.
+
+Unit tests for both scrubbers live in `tests/functional/sentry-scrub.test.ts`.
+
 ## Sample Rates
 
 | Setting | Development | Production |
--- a/instrumentation-client.ts
+++ b/instrumentation-client.ts
@@ -2,11 +2,24 @@ import * as Sentry from "@sentry/nextjs";
 
 Sentry.init({
   dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
-  sendDefaultPii: true,
   tracesSampleRate: process.env.NODE_ENV === "development" ? 1.0 : 0.1,
   replaysSessionSampleRate: 0.1,
   replaysOnErrorSampleRate: 1.0,
-  integrations: [Sentry.replayIntegration()],
+  integrations: [
+    Sentry.replayIntegration({
+      // Mask all text and block all media by default to avoid capturing PII.
+      maskAllText: true,
+      blockAllMedia: true,
+    }),
+  ],
+  beforeSend(event) {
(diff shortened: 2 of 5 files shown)


james-baker:
From Opus xhigh:

PR #84 Review: Remove sendDefaultPii and tighten Sentry replay/header capture

  Overview

  Removes sendDefaultPii: true from both Sentry configs, scrubs auth headers/cookies server-side, masks replay content, and drops client events
  from auth-flow URLs. Scope is appropriate — small, defensive, security-oriented changes aligned with docs/doc-strategy-security.md's posture.

  Code correctness

  - Header deletion is case-sensitive but Sentry normalizes. event.request.headers["authorization"] / ["cookie"] rely on lowercase keys.
  Sentry's Node SDK normalizes header names to lowercase, so this is fine — but worth a one-line comment since a reader might assume HTTP raw
  casing.
  - event.request.cookies covers parsed cookies, but a cookie may also leak via headers["cookie"] — both deletions are needed; the PR correctly
  does both. ✅
  - Client beforeSend URL match is too narrow. /auth/ matches the OAuth callback at src/app/auth/callback/route.ts (good), but the app also has
  /welcome, /invites, and password-reset-style routes that may carry tokens in query strings. Consider also matching event.request?.query_string
   or a broader path list. At minimum, /auth/callback URL fragments (Supabase puts tokens in the URL hash, not query, so they're not sent to
  Sentry — but ?code= is). Worth confirming intent with the author.
  - Dropping all events on /login and /signup is heavy-handed. Real errors on these pages (e.g., a broken form submit, a 500) will now silently
  disappear from Sentry. A safer pattern is to scrub event.request.url/query_string rather than drop the whole event. This is the most material
  concern in the PR.

  Project conventions

  - docs/doc-sentry.md doesn't mention PII handling; consider a short subsection documenting why sendDefaultPii is off and what beforeSend
  strips, per the CLAUDE.md "keep docs in sync" rule.
  - No devjournal entry — this is a security-relevant decision teammates should know about (CLAUDE.md flags this explicitly).

  Test coverage

  - No tests added. The beforeSend functions are pure and trivially unit-testable (pass a fake event, assert headers stripped / event dropped).
  Given the security-sensitive nature, a small Vitest spec under src/ covering both functions would prevent regressions.
  - Test plan in PR body has unchecked boxes — author should run npm test and manually verify per CLAUDE.md "Run all tests before PR."

  Security considerations

  - ✅ Removing sendDefaultPii is the right call.
  - ✅ maskAllText + blockAllMedia is the correct conservative replay default.
  - ⚠️ Server beforeSend doesn't scrub event.request.data (request body) or event.request.query_string. If any route accepts tokens/passwords in
   a body or query, those still leak. Worth at least a comment acknowledging the scope, or scrubbing known sensitive keys.
  - ⚠️ No edge/middleware Sentry config exists, so no parallel change needed — but if one is added later, it needs the same treatment.

  Suggestions

  1. Don't drop entire events on /login//signup — scrub the URL/query_string instead, so genuine errors are still reported.
  2. Add a unit test for both beforeSend callbacks.
  3. Update docs/doc-sentry.md with a "PII handling" subsection.
  4. Consider also scrubbing event.request.query_string server-side, since ?token= style params can appear on auth callback redirects.

  Verdict

  Solid direction, narrow risk. The "drop all events on auth routes" client logic is the one change I'd push back on before merging — it trades
  observability for redundancy with the masking already in place.

james-baker:
Sentry verification checklist

  Setup: wait for the preview deploy on PR #84, sign in, open Sentry → is-app-sentry project in another tab.

  1. Server scrub — cookies & auth headers
  - Hit any authed API route that throws (or temporarily add a throw new Error("scrub-test-server") to one).
  - In Sentry, open the event → Request section.
  - ✅ No cookie or authorization header present.
  - ✅ No cookies block present.
  - ✅ Other headers (user-agent, x-request-id, etc.) still there.

  2. Client scrub — auth-route URL stripping
  - Visit /auth/callback?code=scrub-test-fake (will redirect with an error — that's fine).
  - Trigger a client error on that page (DevTools console: setTimeout(() => { throw new Error("scrub-test-client"); }, 0)).
  - In Sentry, open the event → Request → URL field.
  - ✅ URL ends at /auth/callback, no ?code=....
  - ✅ No query_string field.

  3. Client scrub — non-auth passthrough
  - From / or /invites, trigger another Sentry.captureException.
  - ✅ URL retains its query string (proves we're not over-scrubbing).

  4. Replay masking
  - On any event with a replay attached, open the Replay tab.
  - ✅ Text inputs render as masked blocks (* placeholders).
  - ✅ Images/media blocked or placeholder.

PR #90, opened 2026-04-27
Add --fix flag to check-env.mjs to append missing keys

### LINKED ISSUE #73: Add --fix flag to env preflight that appends missing keys
## Problem
`scripts/check-env.mjs` (added in #60) detects when `.env.local` is missing keys listed in `.env.local.example`, but only tells the dev to copy them over by hand. That's friction — especially when several keys are missing at once or when devs aren't sure exactly which lines to copy.

## Desired outcome
Add a `--fix` flag that appends the missing keys (with their canonical values from `.env.local.example`) to the end of `.env.local`. Update the "missing keys" error message to recommend running with `--fix` instead of manual copy/paste.

## Related
Follow-up to #60 — completes the "non-destructive fix (append missing keys)" outcome described in #70.

NorsemanSpiff:
Closes #73

## Context
Follow-up to the env-preflight work in #60. That PR detects when `.env.local` is missing keys declared in `.env.local.example`, but the only remediation it offers is "copy the lines over manually." This is the "non-destructive fix" outcome described in #70 that #60 deferred.

## What changed
- `scripts/check-env.mjs`: new `--fix` flag. Reads missing keys from `.env.local.example` and appends just the `KEY=VALUE` lines to `.env.local`, prefixed with a dated provenance comment (`# Appended by scripts/check-env.mjs --fix on YYYY-MM-DD`). Existing values stay untouched. Idempotent — a second `--fix` finds nothing missing and exits 0 via the normal success path. Without `--fix`, behavior is unchanged except the error message now recommends running with `--fix` instead of "append manually".
- `scripts/setup.mjs`: hint now points at `--fix` directly so the flag is discoverable from first-time setup output.
- `docs/devjournal.md`: 2026-04-25 entry.

## Why
The more keys we add to `.env.local.example`, the more friction the manual-copy step creates. A dev who hits the preflight error today already has the file open and the keys they need — `--fix` lets them resolve it with one command instead of context-switching to an editor. Existing `.env.local` values (local overrides, custom secrets) are preserved, so the fix is safe to recommend by default.

## Test Plan
- `npm run lint` — clean
- `npm run test:functional` — 7 files / 52 tests pass (after rebase on latest main, which split Vitest into server/client projects)
- Manual: backed up `.env.local`, removed `E2E_ADMIN_PASSWORD` and `CI_RESET_TOKEN`, ran `node scripts/check-env.mjs` (correctly errored, listed both missing keys, recommended `--fix`); ran `node scripts/check-env.mjs --fix` (appended both keys with the dated header, exit 0); re-ran `--fix` (no-op, exit 0); ran plain `check-env` (ok, 7 keys); restored backup.
- Not exercised in CI — CI does `cp .env.local.example .env.local` before any preflight, so `--fix` has no CI surface area.

## Heads-up for reviewer
- Pre-existing typecheck error on `main` at `src/app/page.tsx:84` (`/invites` route literal). Not introduced by this branch — present on bare `origin/main` too — but `npm test` now runs typecheck, so CI may red until that's fixed on main.

### DIFF
--- a/scripts/check-env.mjs
+++ b/scripts/check-env.mjs
@@ -4,9 +4,14 @@
 // dev/test boundary with an actionable error, instead of leaking through to
 // test failures ten layers deep. Keys-only — values are not compared, since
 // devs are free to override locally.
+//
+// Pass --fix to append any missing key lines from .env.local.example to
+// .env.local. Non-destructive: existing values are untouched.
 
-import { existsSync, readFileSync } from "node:fs";
+import { appendFileSync, existsSync, readFileSync } from "node:fs";
 import { resolve } from "node:path";
+
+const fixMode = process.argv.slice(2).includes("--fix");
 
 const repoRoot = resolve(import.meta.dirname, "..");
 const examplePath = resolve(repoRoot, ".env.local.example");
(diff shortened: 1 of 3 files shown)


scripts/setup.mjs
```diff
@@ -13,7 +13,7 @@ function ensureLocalEnv() {
   if (existsSync(target)) {
     console.log("  .env.local already exists — leaving it alone");
     console.log(
-      "  (if tests complain about missing keys, run `node scripts/check-env.mjs`)",
+      "  (if tests complain about missing keys, run `node scripts/check-env.mjs --fix`)",
```
NorsemanSpiff:
```suggestion
      "  (if tests complain about missing keys, run `node scripts/check-env.mjs` to determine what keys are missing, then run `node scripts/check-env.mjs --fix` to append the missing keys from .env.local.example to .env.local)",
```
I imagine there's a shorter way to say this, but would this extra guidance be helpful? The intention is to educate the human that they can run the command without the `--fix` flag by itself to identify which keys .env.local is missing first, then run it with `--fix` to append them.

scripts/check-env.mjs
```diff
@@ -37,20 +42,44 @@ if (!existsSync(localPath)) {
...
+    process.exit(0);
+  }
+
   console.error(
     `check-env: .env.local is missing ${missing.length} key(s) declared in .env.local.example:`,
   );
   for (const key of missing) console.error(`  - ${key}`);
   console.error(
-    "  Fix: append the missing keys from .env.local.example to .env.local",
+    "  Fix: run `node scripts/check-env.mjs --fix` to append the missing keys",
```
NorsemanSpiff:
```suggestion
    "  Fix: run `node scripts/check-env.mjs --fix` to append the missing keys from .env.local.example to .env.local",
```
Would this be a more useful error message?

PR #104, opened 2026-04-30
Add read-only profile view, move edit to /profile/edit

benjifriedman:
## Summary

- `/profile` is now a read-only view of the user's profile, with serif font (Ovo) for user-entered content and sans-serif labels
- Edit form moved to `/profile/edit`, accessible via an "Edit profile" button on the view page
- After saving edits, the form redirects back to `/profile`
- Home page link updated from "Edit profile" to "My profile"
- Replaced all `gray-200` with `gray-500` across the app for better contrast on hover states and text


james-baker:
Looks good to me

PR #111, opened 2026-05-02
Drop VERCEL_ENV gate on /_test/reset

### LINKED ISSUE #108: Add CI_RESET_TOKEN env var to Vercel
E2E tests are failing because the /_test/reset endpoint returns 404 on preview deploys. Since PR #81, that route only registers when CI_RESET_TOKEN is set as an env var at server startup. @james-baker could you add CI_RESET_TOKEN to the Vercel project's environment variables for the Preview environment? The value should match the CI_RESET_TOKEN GitHub Actions secret.

james-baker:
## Summary

- Removes the `VERCEL_ENV !== "production"` check from `/api/_test/reset` so the route registers on production deploys too. `CI_RESET_TOKEN` is now the sole gate.
- Updates `docs/doc-supabase.md` and `docs/plan-security-hardening.md` to reflect the change.

## Why

E2E was failing on `main` and passing on feature branches. `e2e.yml` triggers on every successful `deployment_status` event — including the production deploy after a merge to main — and on production the reset route was absent from the bundle (PR #81 wrapped registration in `isResetEnabled()`). The Playwright setup project 404'd, the suite collapsed, and Benji raised #108 as "add `CI_RESET_TOKEN` to Preview" — but it was already there; the symptom was a missing route on prod, not a missing env var on preview.

Preview and production share one Supabase, so the original environment gate was protecting only the canonical domain's surface, not the data — the token already mutates prod-DB rows on every preview run. Removing it makes the token symmetric across environments, which is the actual security boundary that exists today. The destructive scope is fixed at the two seeded e2e users (`src/server/test-reset.ts`), not arbitrary rows.

Closes #108.

## Manual step on merge

`CI_RESET_TOKEN` needs to be added to the Vercel **Production** environment (it's currently in Preview + Development only). Without it, prod e2e runs will still 404 — they'll just hit the token-mismatch path instead of the route-not-registered path.

## Test plan

- [x] `npm test` passes locally (60 functional + 15 e2e)
- [x] After merge + adding the Vercel Production env var, the e2e run on the main commit succeeds (currently the failing case)
- [x] PR's own preview e2e run still succeeds

### DIFF
--- a/docs/doc-supabase.md
+++ b/docs/doc-supabase.md
@@ -36,18 +36,18 @@ The Playwright e2e suite signs in as two long-lived users instead of admin-provi
 
 Both users should be created with "Auto Confirm User" checked so they can sign in with a password immediately.
 
-**Secrets (set in GitHub Actions → Settings → Secrets → Actions, and as Vercel preview/dev env vars for `CI_RESET_TOKEN`):**
+**Secrets (set in GitHub Actions → Settings → Secrets → Actions, and as a Vercel env var across all environments for `CI_RESET_TOKEN`):**
 
 - `E2E_REGULAR_PASSWORD` — the password for `e2e-regular@testfake.local`. GH Actions only.
 - `E2E_ADMIN_PASSWORD` — the password for `e2e-admin@testfake.local`. GH Actions only.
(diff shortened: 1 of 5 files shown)


james-baker:
https://github.com/Intentional-Society/is-app/commit/7c808e806e1b41aa3a76878737e5ccbe45e1e625 had passing e2e on main.

PR #115, opened 2026-05-03
Add member profile view at /members/[id] (#112)

### LINKED ISSUE #112: See other people's profiles
Feature: User can see other people's profiles (without private (emergency contact) info, without edit affordance)

Clicking on someone in the Directory will go here.

oolu4236:
## Summary
- Implements `getProfileForMember` — public shape that excludes `emergencyContact`, `referredBy`, `referredByLegacy`, and `isAdmin`
- Adds `GET /api/members/:id` endpoint (auth-gated, returns 404 for unknown IDs)
- Creates `/members/[id]` page showing display name, bio, keywords, location, live desire, supplementary info, and member-since year
- Unauthenticated visitors redirected to `/signin`
- Updated functional tests: replaced the old `NotImplemented` stub test with real coverage

## Test plan
- [x] Visit `/members/<uuid-of-seeded-member>` while signed in — profile renders
- [ ] Visit with unknown UUID — 404 page
- [x] Confirm `emergencyContact` is not visible on the page
- [x] Unauthenticated visit redirects to `/signin`

Closes #112.

### DIFF
--- a/src/app/members/[id]/page.tsx
+++ b/src/app/members/[id]/page.tsx
@@ -1,0 +1,59 @@
+import Link from "next/link";
+import { notFound, redirect } from "next/navigation";
+
+import { createClient } from "@/lib/supabase/server";
+import { getProfileForMember } from "@/server/profiles";
+
+function Field({ label, children }: { label: string; children: React.ReactNode }) {
+  return (
+    <div className="flex flex-col gap-1">
+      <dt className="text-sm uppercase tracking-wide text-muted-foreground">{label}</dt>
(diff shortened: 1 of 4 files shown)


james-baker:
I read the code and said to myself, "wait, why is the server component not calling the API?" and then discovered that we have that antipattern in several places already. So, I'm going to merge this and then refactor everything together.

Actually testing /members/foo produces a 500 instead of a 404 or 200. Making that a follow-on defect.

Rendering /members/<uuid> worked, although we'll probably want something prettier than UUID in the URL.

Also, I notice that I want to refactor user auth checking from "duplicated on every page" into proxy.ts or something, but that's not the fault of this commit.

PR #122, opened 2026-05-06
Show friendly message instead of 500 on unknown /members/:id (#116)

### LINKED ISSUE #116: Hitting /members/foo produces 500 error page
Actually testing /members/foo produces a 500 instead of a 404 or 200.

_Originally posted by @james-baker in https://github.com/Intentional-Society/is-app/issues/115#issuecomment-4367830184_

oolu4236:
## Summary
- Replaces `notFound()` with an inline "Member not found" message that stays within the app layout
- Unknown member IDs now show a clear message and a back link instead of the generic Next.js 404 or a 500
- Unexpected non-404 errors still throw so they surface in Sentry

## Test plan
- [ ] Visit `/members/foo` — see "Member not found" with a ← Back link
- [ ] Visit `/members/00000000-0000-0000-0000-000000000001` — see Aria Chen's profile as normal

Closes #116.

### DIFF
--- a/src/app/members/[id]/page.tsx
+++ b/src/app/members/[id]/page.tsx
@@ -1,5 +1,5 @@
 import Link from "next/link";
-import { notFound, redirect } from "next/navigation";
+import { redirect } from "next/navigation";
 
 import { serverApiClient } from "@/lib/api-server";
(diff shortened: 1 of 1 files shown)


james-baker (approved):
LGTM

PR #128, opened 2026-05-07
Add member slugs for readable profile URLs (#127)

### LINKED ISSUE #27: User Directory
* Grid card view 
* Individual profile view. 
* Each card should show avatar, name, keywords, and location

### LINKED ISSUE #127: Member profile URLs use raw UUIDs — add slugs for readable URLs
## Problem

Member profile URLs currently look like `/members/a3b4c5d6-1234-...` — the raw UUID. This is hard to read, hard to share, and surfaced as a concern during review of #115.

Profile IDs are pinned to `auth.users.id` (Supabase-minted UUIDs) so they can't become TypeIDs. A separate `slug` column is the right approach.

## Goal

- Add a `slug` column to `profiles` — auto-generated from display name (e.g. "Aria Chen" → `aria-chen`)
- Route `/members/[slug]` as the canonical URL; fall back to UUID for profiles without a slug
- Directory cards link using the slug when available
- Keep UUID-based routing working so existing links don't break

## Constraints

- Slug must be unique — if two members share a display name, the second keeps a UUID URL until they set a distinct display name
- Profile IDs stay as UUIDs — slugs are display-layer only, not primary keys
- Schema migration required

## Relates to

- #115 (member profile view)
- #27 (member directory)
- `docs/plan-typeid.md` — notes profiles.id can't become a TypeID

oolu4236:
## Summary
- Adds `slug` column to `profiles` (nullable, unique) — Drizzle migration included
- Slug auto-generated from display name on create and update (e.g. "Aria Chen" → `aria-chen`)
- `getProfileForMember` accepts slug OR UUID — both `/members/aria-chen` and `/members/<uuid>` work
- If two members share a display name, the second keeps a UUID URL until they update their name
- Directory cards link by slug when available, UUID otherwise
- Also includes `GET /api/members` endpoint and `/members` directory page (closes #27)
- Home page gains "Member directory" button

## Test plan
- [ ] Visit `/members/aria-chen` — Aria Chen's profile loads
- [ ] Visit `/members/00000000-0000-0000-0000-000000000001` — same profile loads (UUID still works)
- [ ] Visit `/members/does-not-exist` — friendly "Member not found" message
- [ ] Edit display name on `/profile/edit` — slug updates, new slug URL works

Closes #127, closes #27.

### DIFF
--- a/docs/devjournal.md
+++ b/docs/devjournal.md
@@ -4,6 +4,12 @@ Each entry: **Date** | **Author** | **Title**, followed by description text. Mos
 
 ---
 
+## 2026-05-07 | Ola | Readable slug-based profile URLs
+
+Profiles now have readable slug URLs (e.g. `/members/aria-chen` instead of `/members/00000000-...`). Added a nullable unique `slug` column to the profiles schema. Slugs are auto-generated from display name on profile create and update — "Aria Chen" becomes `aria-chen`. `getProfileForMember` accepts either slug or UUID so old UUID links don't break. If two members share a display name the second keeps a UUID URL until they update their name to something distinct. Directory cards link by slug when available, UUID otherwise.
+
+One thing to know: existing profiles in a local DB won't have slugs until you either run `npm run seed:dev` (updated to populate slugs) or trigger a profile save. A Postgres-side backfill on deploy handles production.
+
 ## 2026-05-06 | James | Relations schema: PR 1 of the four-PR Relations ship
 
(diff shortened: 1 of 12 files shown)


james-baker:
I'm processing the merge conflicts since I caused some of them.

I'm also reverting some of the db-catch code, you saw a real problem (prod db isn't migrated when PR is doing e2e against preview deployment) but there's a different pattern (forward deploy non-breaking migration to prod db) to solve it.

james-baker:
One note, though, @oolu4236  re this PR - it had the changes of your previous PR _also_ submitted as a part of _this_ PR. After you create a PR, if you don't want to merge it but you want to do other work, I think you need to "stash" that code locally in order to get another separate clean change (if you want to do something else that's not building on top of it).

Issue #131, opened 2026-05-07
Make profile pictures work


james-baker:
Pick a backend solution (supabase filestore? blob in database? S3 bucket and CDN?)

Code for profile-editing that lets people upload a profile photo.

Pick a little UI library for cropping/management of profile photos?

Then find some code pattern for using photos:
Do they have an img URL that is a part of their profile? (which gets created at file upload time?)
Or is there some other "fetch photo" code idiom between nextjs and supabase?

james-baker:
From #136 (closing as dup): avatar URLs are currently user-supplied freetext rendered via plain `<img>`, which means any HTTPS host is fetched by every visitor (tracking-pixel risk) and we can't use `next/image` optimisation. The fix lands with this work — upload to Supabase Storage, serve via `next/image` with `*.supabase.co` whitelisted in `images.remotePatterns`, and drop or restrict the freetext `avatarUrl` field.

🤖 Filed by [Claude Code](https://claude.com/claude-code)

PR #135, opened 2026-05-08
Polish member directory

AlexisChen99:
## Summary

- Switch member directory from a single-column list to a responsive 4-column grid (1 col mobile → 2 col tablet → 4 col desktop)
- Add avatar image to each member card with an initials fallback for members without a photo
- Render keywords as pill badges to visually distinguish them from the location field
- Add `avatarUrl` to `MemberSummary` type and `listMembers` query
- Add 4 new local dev seed members (IDs 0016–0019) — no effect on production

## Test plan

- [ ] Visit `/members` on the Vercel preview and confirm 4-column grid renders on desktop
- [ ] Confirm 1-column layout on mobile
- [ ] Confirm initials placeholder shows for members without an avatar
- [ ] Confirm keywords appear as pills, location as plain text


AlexisChen99:
Screenshot 2026-05-08 at 12 36 42 AM

oolu4236 (commented):
Good improvements overall — the grid layout, initials fallback, and keyword pills all make the directory feel much more like a real product.

One thing worth flagging: the `` tag loading `avatarUrl` directly. The `biome-ignore` comment is correct that `next/image` requires whitelisted domains, but the underlying concern is real — any member can point their avatar at an arbitrary URL, which means:
- External requests from every visitor's browser to potentially untrusted hosts
- Tracking pixel risk
- The CSP `img-src 'self' data: blob: https:` we have allows any HTTPS image source, so it won't catch this

Since avatar upload isn't built yet and members are manually entering URLs, the blast radius is low for now — it's the same trust level as `avatarUrl` in the profile edit form. But worth tracking.

When avatar upload lands, the right fix is either:
- Upload to Supabase Storage and whitelist `*.supabase.co` in `next.config.ts` `images.remotePatterns`, then use `<Image>`
- Or proxy through our own endpoint so the external request never comes from the visitor's browser

Not a blocker for this PR given the current state of avatar support. Just want it on the record before we add upload.

oolu4236 (approved):
Looks good to merge. Grid layout, avatar initials, and keyword pills are all solid improvements. Left a comment on the img tag / CSP tradeoff for when avatar upload lands — not a blocker.

james-baker:
The (AI?) concerns about profile picture `s` should be handled by #131

james-baker (approved):
LGTM

Issue #149, opened 2026-05-11
flaky db transactions using transaction-level connection pooler (was: Playwright e2e failures with 20s timeouts)


james-baker:
## Context

Recent e2e runs against Vercel preview URLs are failing on the **first** `completeWelcome` call in each browser-context cycle. The latest `main` run ([25654953740](https://github.com/Intentional-Society/is-app/actions/runs/25654953740) on `930fd57`) had three failures:

- `/signup — unauthed invite flow › valid code → note displayed → submitting email shows 'check your email'`
- `invites — authed member flow › create an invite, see it listed, revoke it`
- `/myweb — first-time welcome tour › tour appears for a first-time visitor and can be skipped`

All three time out the same way:

```
TimeoutError: page.waitForURL: Timeout 20000ms exceeded.
> 59 |   await page.waitForURL((url) => url.pathname === "/welcome", {
```

inside `tests/e2e/helpers/session.ts`'s `completeWelcome` helper. The 20s limit replaced an older 10s value just before this issue surfaced (commit `8af53b0`) and is now also being exhausted.

## What we know

`signInAs` succeeds — its own `waitForURL` clearing `/signin` passes cleanly. The 20s budget runs out inside `completeWelcome` waiting for the post-signin chain (`/signin → / → middleware → /welcome`) to settle.

The failure pattern is consistent: **always the first** `completeWelcome` per browser-context cycle. Sibling tests reusing the same `beforeEach` and helper pass cleanly once the preview has been hit at least once:

- `invites — authed member flow › 429 from the API surfaces a friendly cap message` (shares `beforeEach` with the failing test — passes).
- `/myweb — tour pre-dismissed › Done/Edit toggle` (calls `completeWelcome` too — passes).

So the cost is a per-instance, one-time-only thing. The pattern points at *something* that runs on the first request to a fresh deployment and not again.

Trend across runs:

| Commit | Failures |
| --- | --- |
| PR supabase/supabase#148 build (`d2e619c`) | 1 — `/signup — unauthed invite flow` only |
| PR supabase/supabase#147 build | 1–2 (same test + intermittent member-flow) |
| `main` `930fd57` | 3 — all `completeWelcome`-first tests |

Each redeploy widens the failing surface by one test.

## What we don't know

[...]

PR #151, opened 2026-05-11
Drop Sentry `includeLocalVariables` to cut server cold-start

james-baker:
## Summary

Drops the `includeLocalVariables: true` option from `sentry.server.config.ts`. Sentry implements the "local variables in stack frames" feature via the V8 inspector — it attaches at module load and stays attached, intercepting every function call to keep frame info available for error capture. That's a real cold-start cost and a steady-state runtime tax on every server request.

Trade-off: less context in captured server-side error reports. Re-enable per environment if a specific debugging session needs it.

This is option (A) from #149's investigation plan — the cheapest lever first, before considering the bigger structural question of how often we hit `supabase.auth.getUser()` on every request.

Investigates #149 (not closing — landing this lets us observe whether the e2e cold-start flakes change before deciding on next steps).

## Test plan

- [x] `npm run test:functional` — 137 / 138 pass (the 1 skipped is pre-existing).
- [ ] CI lint + functional tests.
- [ ] Vercel preview build + Playwright e2e — primary signal we're watching, since it's the cold-start path that triggered #149.

### DIFF
--- a/sentry.server.config.ts
+++ b/sentry.server.config.ts
@@ -5,6 +5,11 @@ import { scrubServerEvent } from "@/lib/sentry-scrub";
 Sentry.init({
   dsn: process.env.SENTRY_DSN,
   tracesSampleRate: process.env.NODE_ENV === "development" ? 1.0 : 0.1,
-  includeLocalVariables: true,
+  // includeLocalVariables is intentionally omitted: it attaches the V8
+  // inspector at module load and intercepts every function call to keep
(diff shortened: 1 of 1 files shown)


james-baker:
Closing — the retry-failure pattern (try 1 + 2 retries all timing out at 20s on warm preview) rules out cold-start as the dominant cost, so this change isn't justified by #149. We never measured an independent impact for the Sentry config either, so shipping it on speculation isn't worth it.

Issue #141, opened 2026-05-09
Make script to import our existing user data from csv


james-baker:
Depends on #131 happening first

This is to take our google sheet signup data (and the photos uploaded by google forms to google drive) and get all that transferred into is-app.

This could be a one-time thing (if we're ready to retire the google form), or it could be run multiple times if we want to keep using the google form for signup during a testing-with-existing-users period.

james-baker:
## Prod run procedure (after PR #221 merges)

1. Merge PR #221.
2. Create a temporary `.env.prod` (gitignored via `.env.*`) with the production `NEXT_PUBLIC_SUPABASE_URL`, `SUPABASE_SECRET_KEY`, and the pooler `DATABASE_URL`.
3. Import — `--overwrite` so the dev-team profiles already on prod are aligned to the CSV; you'll be asked to type the host to confirm:
   ```
   npx tsx scripts/import-members-csv.ts "scripts/IS Web Form Responses Master - Form Responses 1.csv" --prod --overwrite
   ```
4. Generate referral links — `npx tsx scripts/normalize-referrals.ts --prod`, then review the SQL it writes under `scripts/.import-logs/`.
5. Apply the reviewed file — `npx tsx scripts/normalize-referrals.ts --apply <file.sql> --prod`.
6. Delete `.env.prod`.

Each importer run also writes a timestamped log to `scripts/.import-logs/`.

## ⚠️ Preserve `scripts/.import-cache/`

The photo cache holds the two hand-converted HEIC photos that `sharp` can't decode — file IDs `1PN3_j9nJRIfd7d2LWMJD747QzF-tqdJT` and `1nyYbqiw_eKju_rWp5vowkTI5LFlUamNr`. The same folder will be reused for the prod run. Do **not** delete it — losing it means re-converting those two photos by hand.

---
🤖 Posted by Claude Code

PR #182, opened 2026-05-17
Add CSV member import script (#141)

oolu4236:
## Summary
- New `scripts/import-members-csv.mjs` — reads a CSV export from the IS Google Sheet signup form and upserts members into Supabase
- Auth users created via Admin API (`email_confirm: true`, no password — members sign in via magic link)
- Profile rows use `INSERT ... ON CONFLICT DO UPDATE` so re-running is safe
- `--dry-run` flag previews without writing anything
- Column mapping is configurable at the top of the file to match actual CSV headers
- `npm run import:members <path/to/csv> [--dry-run]`

## What it does NOT handle
- Photo import (Google Drive → Supabase Storage) — out of scope, tracked separately
- The `avatarUrl` mapping is set to `null` by default but can be pointed at a CSV column with public URLs

## Test plan
- [ ] Run with `--dry-run` against a real CSV export — confirm row count and emails look correct
- [ ] Run against local dev (`npm run dev:db`) to confirm users and profiles are created
- [ ] Re-run against the same CSV — confirm no duplicates, existing profiles updated

EDIT: Does not close the issue, needs further work

### DIFF
--- a/scripts/import-members-csv.mjs
+++ b/scripts/import-members-csv.mjs
@@ -1,0 +1,245 @@
+#!/usr/bin/env node
+/**
+ * Import IS member data from a CSV export of the Google Sheet signup form.
+ *
+ * Usage:
+ *   node scripts/import-members-csv.mjs path/to/members.csv
+ *   node scripts/import-members-csv.mjs path/to/members.csv --dry-run
+ *
+ * Required env vars (same as the app):
+ *   NEXT_PUBLIC_SUPABASE_URL
+ *   SUPABASE_SECRET_KEY
+ *   DATABASE_URL
+ *
+ * Column mapping:
+ *   Edit COLUMN_MAP below to match your CSV headers. Keys are profile
+ *   field names; values are the exact CSV column header strings.
+ *   Set a value to null to skip that field.
+ *
+ * Idempotency:
+ *   - Auth users are created with createUser if they don't exist, or
+ *     looked up by email if they do. Re-running is safe.
+ *   - Profile rows use INSERT ... ON CONFLICT DO UPDATE so edits to
+ *     the CSV and re-running will overwrite previous imports.
+ *   - Rows with no email are skipped with a warning.
+ *
+ * Photos:
+ *   Google Form saves photos to Google Drive. Downloading and uploading
+ *   them to Supabase Storage requires the Google Drive API and is not
+ *   handled here. See docs/doc-email.md for the planned approach.
+ *   The avatarUrl column in the CSV (if present) is imported as-is
+ *   and should be a publicly accessible URL or a Supabase Storage path.
+ */
+
+import { createReadStream } from "node:fs";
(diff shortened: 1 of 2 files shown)


james-baker:
@oolu4236 What is the exact procedure you're using to produce the CSV file?

scripts/import-members-csv.mjs
```diff
@@ -0,0 +1,245 @@
...
+ *   - Auth users are created with createUser if they don't exist, or
+ *     looked up by email if they do. Re-running is safe.
+ *   - Profile rows use INSERT ... ON CONFLICT DO UPDATE so edits to
+ *     the CSV and re-running will overwrite previous imports.
+ *   - Rows with no email are skipped with a warning.
+ *
+ * Photos:
+ *   Google Form saves photos to Google Drive. Downloading and uploading
+ *   them to Supabase Storage requires the Google Drive API and is not
+ *   handled here. See docs/doc-email.md for the planned approach.
```
james-baker: I find this doc has nothing to do with GDrive or even profile photos. Hallucination?

scripts/import-members-csv.mjs
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * Column mapping:
+ *   Edit COLUMN_MAP below to match your CSV headers. Keys are profile
+ *   field names; values are the exact CSV column header strings.
+ *   Set a value to null to skip that field.
+ *
+ * Idempotency:
+ *   - Auth users are created with createUser if they don't exist, or
+ *     looked up by email if they do. Re-running is safe.
+ *   - Profile rows use INSERT ... ON CONFLICT DO UPDATE so edits to
```
james-baker: Yikes, probably the opposite of what we want (because app edits are presumably newer than the CSV data). We're not going to be editing the CSV file (...other than maybe adding rows?)

oolu4236:
Addressed both review points:
- Changed profile upsert to `ON CONFLICT DO NOTHING` — app edits take precedence over CSV data, re-running won't overwrite member changes
- Removed the incorrect reference to `docs/doc-email.md` (had nothing to do with Google Drive photos — hallucination)

james-baker:
I'm gonna pick this (#141) up personally because I want to be extra careful with people's PII and with prod data. And we totally need a solution for the photos also - we can't just declare it "out of scope", and "tracked separately" seems like a hallucination 'cause I don't see it.

james-baker:
> @oolu4236 What is the exact procedure you're using to produce the CSV file?

Still would like a response to this

james-baker:
> > @oolu4236 What is the exact procedure you're using to produce the CSV file?
> 
> Still would like a response to this

I've done a straightforward "Download --> CSV" from the GSheet UI and am going to bake in a filename "IS Web Form Responses Master - Form Responses 1.csv" for this basically-one-off script which currently takes a filename input.

Dropping my question.

PR #184, opened 2026-05-17
Replace bio-null welcome gate with explicit lastUpdatedProfile (#183)

### LINKED ISSUE #183: Welcome redirect heuristic (bio === null) is fragile
## Problem

The logged-in home page redirects members to `/welcome` if `profile.bio === null`:

```ts
if (me.profile && me.profile.bio === null) {
  redirect("/welcome");
}
```

This breaks in two real scenarios:
1. A member who completes welcome but later clears their bio from `/profile/edit` gets permanently redirected back to `/welcome`
2. A future change to what the welcome form collects (e.g. bio becomes optional) silently breaks the gate for all members

## Goal

Replace the bio sentinel with an explicit `welcomeCompletedAt` timestamp column on `profiles`. Set it once when the welcome form is successfully submitted. The redirect gate becomes `welcomeCompletedAt IS NULL`.

## Implementation

1. Add `welcome_completed_at timestamptz` column to `profiles` (nullable, no default)
2. Generate and apply migration
3. Set `welcomeCompletedAt = now()` in the `PUT /me` handler when called from the welcome form — or add a dedicated `POST /api/me/complete-welcome` endpoint
4. Update the redirect gate in `src/app/page.tsx` from `bio === null` to `welcomeCompletedAt === null`
5. Update `resetE2EUsers` in `test-reset.ts` to also clear `welcomeCompletedAt`

## Notes

- Existing members (bio already set) need `welcomeCompletedAt` backfilled — a one-time `UPDATE profiles SET welcome_completed_at = now() WHERE bio IS NOT NULL` covers this
- The bio field itself stays as-is — this is only about the gate

oolu4236:
## Problem

`bio === null` was a fragile sentinel for whether a member had completed `/welcome`. A member who cleared their bio after onboarding was trapped in a permanent redirect loop back to `/welcome`.

## Summary

- Adds `last_updated_profile timestamptz` (nullable) to `profiles` — migration `0005`.
- `PUT /me` sets `lastUpdatedProfile = now()` on every profile save.
- The gate in `page.tsx` redirects to `/welcome` whenever `lastUpdatedProfile` is null — i.e. until the member has saved their profile at least once.
- `test-reset.ts` clears `lastUpdatedProfile` alongside the other fields so E2E runs start clean.

## Behaviour for existing members

A member who has never saved their profile through the app has `lastUpdatedProfile = null` and is routed through `/welcome` once. This is intended: it covers CSV-imported members with pre-filled profiles, whom we want to confirm and update their details. Saving the welcome form issues a `PUT /me`, which sets the timestamp and clears the gate permanently.

## Schema rollout

`0005` is additive (a nullable `ADD COLUMN`), so it is applied to the production database ahead of merge via `npm run prod:db:expand`. This keeps the preview's E2E green: previews share the prod DB and skip migrations, so the column must already exist for the reset endpoint to succeed. See the expand-contract section of `docs/strategy-committing.md`.

## Test plan

- [ ] New member completes welcome → lands on home, `lastUpdatedProfile` is set
- [ ] Existing member who has never saved their profile → routed to `/welcome` once
- [ ] After saving the welcome form → no longer redirected, even if bio is later cleared

Closes #183.

### DIFF
--- a/drizzle/0005_certain_omega_flight.sql
+++ b/drizzle/0005_certain_omega_flight.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "profiles" ADD COLUMN "welcome_completed_at" timestamp with time zone;
--- a/drizzle/0005_certain_omega_flight.sql
+++ b/drizzle/0005_certain_omega_flight.sql
@@ -1,1 +1,0 @@
-ALTER TABLE "profiles" ADD COLUMN "welcome_completed_at" timestamp with time zone;
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -99,10 +99,11 @@ export default async function Home() {
     );
   }
 
-  // Incomplete-profile heuristic: bio null means the member has not
-  // completed /welcome yet. bio is the one field the welcome form
-  // always collects, so it's a reliable sentinel.
-  if (me.profile && me.profile.bio === null) {
+  // Gate: redirect to /welcome until the member has explicitly submitted
+  // the welcome form (welcomeCompletedAt set by POST /api/me/complete-welcome).
+  // Falls back to the old bio-null heuristic for members who completed
+  // welcome before this column existed, so existing members aren't re-gated.
+  if (me.profile && !me.profile.welcomeCompletedAt && me.profile.bio === null) {
     redirect("/welcome");
   }
 
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -99,11 +99,11 @@ export default async function Home() {
     );
   }
 
-  // Gate: redirect to /welcome until the member has explicitly submitted
-  // the welcome form (welcomeCompletedAt set by POST /api/me/complete-welcome).
-  // Falls back to the old bio-null heuristic for members who completed
-  // welcome before this column existed, so existing members aren't re-gated.
-  if (me.profile && !me.profile.welcomeCompletedAt && me.profile.bio === null) {
+  // Gate: redirect to /welcome until the member has saved their profile
+  // at least once (lastUpdatedProfile set by PUT /me). This includes
+  // pre-filled profiles from the CSV import — we want those members to
+  // go through the welcome flow to confirm and update their details.
+  if (me.profile && !me.profile.lastUpdatedProfile) {
     redirect("/welcome");
   }
 
--- a/src/app/welcome/welcome-form.tsx
+++ b/src/app/welcome/welcome-form.tsx
@@ -7,6 +7,7 @@ import { ProfileFields } from "@/components/profile-fields";
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
 import { Label } from "@/components/ui/label";
+import { apiClient } from "@/lib/api";
 import { createClient } from "@/lib/supabase/client";
 import { type ProfileFormValues, useProfileForm } from "@/lib/use-profile-form";
 
@@ -31,6 +32,8 @@ export function WelcomeForm({ initial }: { initial: ProfileFormValues }) {
         return;
       }
     }
+
+    await apiClient.api.me["complete-welcome"].$post();
 
     router.push("/");
     router.refresh();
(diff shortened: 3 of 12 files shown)


src/server/profiles.ts
```diff
@@ -98,6 +98,7 @@ export type ProfileForSelf = {
   emergencyContact: string | null;
   liveDesire: string | null;
   isAdmin: boolean;
+  welcomeCompletedAt: Date | null;
```
james-baker: let's make this lastUpdatedProfile instead

src/app/welcome/welcome-form.tsx
```diff
@@ -77,6 +77,8 @@ export function WelcomeForm({ initial }: { initial: Initial }) {
         }
       }
 
+      await apiClient.api.me["complete-welcome"].$post();
```
james-baker: check for res.ok and surface error? - nevermind, let's remove this endpoint instead.

src/server/api.ts
```diff
@@ -112,6 +113,11 @@ const api = new Hono<{ Variables: ApiVariables }>()
     const profile = await getProfileForSelf(user.id);
     return c.json({ id: user.id, email: user.email, profile });
   })
+  .post("/me/complete-welcome", async (c) => {
```
james-baker: Let's make the PUT on "me" set the lastUpdatedProfile value instead of making a separate call.

drizzle/0005_certain_omega_flight.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "profiles" ADD COLUMN "welcome_completed_at" timestamp with time zone;
```
james-baker: When this is revised and the PR is ready to go again, run "npm run prod:db:expand" to submit the nonbreaking/expansionary schema change for forward-deployment

src/app/page.tsx
```diff
@@ -119,10 +119,11 @@ export default async function Home() {
...
 
-  // Incomplete-profile heuristic: bio null means the member has not
-  // completed /welcome yet. bio is the one field the welcome form
-  // always collects, so it's a reliable sentinel.
-  if (me.profile && me.profile.bio === null) {
+  // Gate: redirect to /welcome until the member has explicitly submitted
+  // the welcome form (welcomeCompletedAt set by POST /api/me/complete-welcome).
+  // Falls back to the old bio-null heuristic for members who completed
+  // welcome before this column existed, so existing members aren't re-gated.
+  if (me.profile && !me.profile.welcomeCompletedAt && me.profile.bio === null) {
```
james-baker: Take away the fallback: If we've prefilled their profile (with stuff from the google form), we want to take them through the "here's where/how you update your profile" path even if their bio is not null.

james-baker (changes requested):
Changing to "lastUpdatedProfile" is the shift from which the other changes flow. (180 landed first, but the merge conflict might disappear after the updates.)

oolu4236:
Addressed all review points:
- Renamed `welcomeCompletedAt` → `lastUpdatedProfile` (column, type, migration, all usages)
- Removed `POST /api/me/complete-welcome` endpoint — `PUT /me` now sets `lastUpdatedProfile = now()` on every save, no extra call needed
- Removed the bio-null fallback from `page.tsx` — redirect to /welcome fires solely when `lastUpdatedProfile` is null (this includes CSV-imported members with pre-filled bios, so they go through the welcome flow to confirm their details)
- Migration regenerated as `drizzle/0005_add_last_updated_profile.sql`

drizzle/0005_certain_omega_flight.sql
james-baker: I'm looking at the github Actions tab and prod:db:expand has not yet been run. I want to check, do you understand this process? After we deploy the schema expansion, then the e2e will run green and we can merge the code changes to main. This wasn't well-specified before - I have updated the strategy-committing doc to integrate this script deploy-to-db process into the documented workflow pattern. in #214

james-baker (changes requested):
just needs a db deployment before checks will run green

james-baker:
gonna finish this, with schema expansion, myself now.

james-baker (approved):

PR #190, opened 2026-05-17
Add search/filter to member directory (#186)

### LINKED ISSUE #186: Member directory needs search/filter
## Problem

The member directory at `/members` lists everyone alphabetically. With a growing membership this becomes unusable — there is no way to filter by keyword, location, or name.

## Goals

- Add a search/filter bar to `/members` that filters the visible cards client-side
- Filter should match against: display name, location, and keywords
- No backend change needed — all member data is already loaded on the page
- Should be instant (no debounce needed for a list of this size)
- Empty state: "No members match your search" when nothing matches

## Implementation notes

- Client component wrapping the existing server-rendered list, or convert the page to fetch client-side
- The existing `MemberCard` and grid layout stay unchanged
- Consider a single text input rather than separate field filters for v1

oolu4236:
## Summary
- Extracts the member grid into a `MembersSearch` client component with an instant text filter
- Filters by display name, location, and keywords as you type — no debounce needed at this scale
- Empty state: "No members match your search." when query returns nothing
- Server page stays server-rendered for the initial data fetch; only the filter is a client component

## Test plan
- [ ] Visit `/members`, type a name — cards filter instantly
- [ ] Type a keyword — cards with that keyword remain
- [ ] Type a location — cards with that location remain  
- [ ] Clear input — all cards return
- [ ] No results — "No members match your search." shown

Closes #186.

### DIFF
--- a/src/app/members/members-list.tsx
+++ b/src/app/members/members-list.tsx
@@ -31,7 +31,7 @@ function MemberCard({ member }: { member: MemberSummary }) {
   );
 }
 
-export function MembersSearch({ members }: { members: MemberSummary[] }) {
+export function MembersList({ members }: { members: MemberSummary[] }) {
   const [query, setQuery] = useState("");
 
   const filtered = query.trim()
--- a/src/app/members/members-search.tsx
+++ b/src/app/members/members-search.tsx
@@ -1,0 +1,73 @@
+"use client";
+
+import { useState } from "react";
+import type { UrlObject } from "node:url";
+import Link from "next/link";
+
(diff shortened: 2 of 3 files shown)


james-baker (changes requested):
MembersSearch should be called MembersList (it has a lot more than just search in it, and "List" parallels "ProgramsList"). Otherwise looks good!

oolu4236:
Renamed `MembersSearch` → `MembersList` (component and file) as requested.

james-baker (approved):

PR #191, opened 2026-05-17
Freeze slug on first set — never update on display name change (#188)

### LINKED ISSUE #188: Changing display name silently breaks old slug URLs
## Problem

Member profile slugs are derived from display names (e.g. "Aria Chen" → `/members/aria-chen`). When a member changes their display name, the slug is updated — but any existing links to the old slug (`/members/aria-chen`) return a 404 with no redirect.

This affects shared links, bookmarks, and any future internal links to member profiles.

## Goal

When a member's slug changes, redirect the old slug to the new one.

## Options

1. **Redirect table** — store `(old_slug, new_slug, profile_id)` rows and resolve them in `getProfileForMember`. Clean but requires a migration and extra query.
2. **Keep old slug** — never change the slug once set, even if display name changes. Simpler but means slugs drift from names.
3. **Soft redirect via profile_id** — if slug lookup fails, try UUID lookup and redirect to the current slug URL. Requires passing the UUID around.

Option 2 is probably right for v1: slugs are set once on first save and never updated, keeping them stable. The display name and slug can diverge — that's fine. Document this as intentional.

## Notes

- Currently `PUT /me` updates slug whenever displayName changes (via `toSlug()` in `profiles.ts`)
- Option 2 would mean removing the slug update from `PUT /me` — only set on first upsert

oolu4236:
## Summary
Removes the slug update from `PUT /me`. Slugs are set once by `upsertProfile` at signup and never changed, keeping `/members/aria-chen`-style URLs stable even if the member later edits their display name.

The slug update in `PUT /me` silently broke existing links whenever a member changed their display name — the old URL would 404 with no redirect.

## Trade-off
Slugs can now drift from display names (e.g. a member who changes from "Aria Chen" to "Aria Smith" keeps the slug `aria-chen`). This is acceptable: slugs are identifiers, not mirrors of the current name. A member wanting a new slug can contact an admin, or we can add self-service slug editing later.

## Test plan
- [ ] Edit display name in `/profile/edit` — slug does not change
- [ ] Old profile URL still resolves after display name change

Closes #188.

### DIFF
--- a/src/server/api.ts
+++ b/src/server/api.ts
@@ -16,7 +16,6 @@ import {
   listMembers,
   markWebUpdated,
   parseEditableProfile,
-  toSlug,
   upsertProfile,
 } from "./profiles";
 import { joinProgram, leaveProgram, listPrograms } from "./programs";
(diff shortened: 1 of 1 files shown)


james-baker:
This trades one problem (an old link using an old name would 404 without redirect) for another (changing their name leaves their old name in the url), and I like the first problem better. I am a fan of "add self-service slug editing to the profile-editing path" for later.

PR #192, opened 2026-05-17
Add account deletion with confirmation UI (#185)

### LINKED ISSUE #185: No way to delete your account
## Problem

Members can create an account, upload a photo, fill out a profile, join programs, and build a relational web — but there is no way to leave. There is no account deletion flow anywhere in the app.

This is a GDPR requirement (right to erasure) and a basic expectation for any web app that stores personal data.

## Goals

- Allow a signed-in member to permanently delete their account from the profile settings
- Deletion should remove: auth.users row, profiles row, avatar from Storage, relations, program memberships, and optionally invites created by the member
- Show a confirmation step before deleting (irreversible action)
- Sign the member out and redirect to the logged-out home page after deletion
- Admin users should not be able to self-delete without first revoking their admin flag (guard against orphaned admin-only state)

## Notes

- Requires `SUPABASE_SECRET_KEY` on the server to call `admin.auth.admin.deleteUser()`
- The `avatars` bucket cleanup should use the existing `clearAvatar` helper
- Consider a 30-day soft-delete / grace period in the future, but hard delete is acceptable for v1

oolu4236:
## Summary
- `DELETE /api/me` endpoint: clears avatar from Storage → deletes profile row (FK cascades handle program memberships, relations, invite_hints) → deletes `auth.users` row via Admin API
- Admin accounts return 403 with `admins_cannot_self_delete` — they must revoke their own admin flag first
- `/profile` shows a "Delete my account" link at the bottom, separated by a divider
- Clicking reveals a confirmation panel with a destructive button and cancel
- After deletion: redirects to `/` which bounces to `/signin` since the session is gone

## What cascades automatically
- `profile_programs` — cascade delete
- `relations` (both directions) — cascade delete  
- `invite_hints` — cascade delete
- `invites.created_by` — set null (invites stay, creator becomes anonymous)

## Test plan
- [ ] Sign in, go to `/profile`, click "Delete my account" — confirmation panel appears
- [ ] Click Cancel — panel disappears, no deletion
- [ ] Click "Yes, delete my account" — redirected to `/signin`
- [ ] Confirm user no longer exists in Supabase dashboard
- [ ] Confirm avatar removed from Storage bucket
- [ ] Admin account: confirm 403 error message shown

Closes #185.

### DIFF
--- a/src/app/profile/delete-account-button.tsx
+++ b/src/app/profile/delete-account-button.tsx
@@ -1,0 +1,76 @@
+"use client";
+
+import { useRouter } from "next/navigation";
+import { useState } from "react";
+
+import { apiClient } from "@/lib/api";
+import { Button } from "@/components/ui/button";
+
+export function DeleteAccountButton() {
+  const router = useRouter();
+  const [confirming, setConfirming] = useState(false);
+  const [deleting, setDeleting] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+
+  const handleDelete = async () => {
+    setDeleting(true);
+    setError(null);
+    try {
+      const res = await apiClient.api.me.$delete();
+      if (!res.ok) {
+        const body = (await res.json().catch(() => ({}))) as { error?: string };
+        if (body.error === "admins_cannot_self_delete") {
+          setError("Admin accounts cannot be self-deleted. Ask another admin to remove your admin flag first.");
+        } else {
+          setError("Something went wrong. Please try again.");
+        }
+        setDeleting(false);
+        return;
+      }
+      router.push("/");
+      router.refresh();
+    } catch {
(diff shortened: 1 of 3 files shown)


james-baker:
I'm gonna wait on this one. When ready, we need to:
* decide how much of a person's existence to delete (maybe we keep their name on the ghost?)
* review the "ON DELETE"s wired throughout the DB tables, which should be CASCADE or other values
* explain on the "delete my account" confirmation what will and won't be removed (i.e. other people's memory-data about their history won't get deleted)
* figure out any delta between our regular delete and a GDPR "right to be forgotten" deletion (which could be a manual process upon request)

james-baker:
this code is small enough (and old enough now that), let's just remake it when we want to tackle #185.

PR #200, opened 2026-05-17
Add change password section to profile edit page (#194)

### LINKED ISSUE #194: No way to change password from profile settings
## Problem
The optional password field only exists in the welcome form (`/welcome`). After onboarding, there is no way to change or set a password from `/profile/edit`. A member who skipped setting a password during welcome, or who wants to change their password later, has no path to do so.

## Goal
Add a "Change password" section to `/profile/edit` (or a separate `/profile/security` page) where members can set or update their password.

## Implementation
- Add a password field (and confirm password field) to `/profile/edit` below the main profile fields
- On save, call `supabase.auth.updateUser({ password })` if the field is non-empty
- Validate that new password meets minimum length before submitting
- Show success/error feedback inline

oolu4236:
## Summary
- New `ChangePasswordForm` component on `/profile/edit` below the main profile fields
- Validates min length (8 chars) and password match client-side before submitting
- Calls `supabase.auth.updateUser({ password })` on success
- Shows "Password updated." or an error message inline
- No backend changes needed — uses existing Supabase client auth

## Test plan
- [ ] Go to `/profile/edit` — "Change password" section visible below profile fields
- [ ] Submit mismatched passwords — error shown, no API call
- [ ] Submit < 8 chars — error shown
- [ ] Submit valid matching passwords — "Password updated." shown
- [ ] Sign out and back in with new password — works

Closes #194.

### DIFF
--- a/src/app/profile/change-password-form.tsx
+++ b/src/app/profile/change-password-form.tsx
@@ -26,13 +26,19 @@ export function ChangePasswordForm() {
     }
     setStatus({ kind: "submitting" });
     const supabase = createClient();
-    const { error } = await supabase.auth.updateUser({ password });
-    if (error) {
-      setStatus({ kind: "error", message: error.message });
-    } else {
-      setStatus({ kind: "success" });
-      setPassword("");
-      setConfirm("");
+    try {
+      const { error } = await supabase.auth.updateUser({ password });
+      if (error) {
+        setStatus({ kind: "error", message: error.message });
+      } else {
+        setStatus({ kind: "success" });
+        setPassword("");
+        setConfirm("");
+      }
+    } catch {
+      // A thrown error (network drop, CORS) bypasses the returned { error };
+      // catch it so the button doesn't stick on "Saving…" with no feedback.
+      setStatus({ kind: "error", message: "Something went wrong. Please try again." });
     }
   };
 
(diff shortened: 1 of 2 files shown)


james-baker (changes requested):
Requesting changes — the change-password section is close, but it needs copy fixes, one bug fix, and a follow-up split out.

## Copy / labels — in this PR

1. Section heading (`edit/page.tsx`): **"Change password" → "Set or change password"** — magic-link-only members have no password yet, so "change" alone undersells it.

2. Helptext (`edit/page.tsx`): replace *"Leave blank if you prefer to sign in with magic links."* with **"If you prefer signing in via email, you don't need a password."** The current line implies the password rides along with the profile save — it doesn't, this is a separate form with its own submit — and "leave blank" doesn't match the `required` inputs.

3. Submit button (`change-password-form.tsx`): idle label **"Update password" → "Set/update password"**. Leave the "Saving…" state unchanged.

## Bug fix — in this PR

4. Guard the `updateUser` call against throws. `handleSubmit` awaits `supabase.auth.updateUser({ password })` with no try/catch. It handles the returned `{ error }`, but an actual throw (network drop, CORS) goes unhandled — `status` stays `submitting` and the button sticks on "Saving…" with no feedback. Wrap it and surface the error. Same defect tracked in #211 for the welcome form; this is a separate copy of it.

## Follow-up — "Remove password" — separate PR, not this one

A "stop using a password" path is worth having (let a member who set a password drop back to email-only sign-in), but it's a backend feature rather than a label tweak, so it shouldn't block #200. Filed as #212 — the feasibility is thornier than it looks (Supabase has no API to clear a password, and no reliable way to detect whether one is set); see the issue for details.

🤖 Review by [Claude Code](https://claude.com/claude-code)

james-baker (approved):
implemented changes

PR #208, opened 2026-05-17
Copy full signup link instead of bare invite code (#202)

### LINKED ISSUE #202: Invites panel copies code only — should copy full signup link
## Problem
The invites panel has a "Copy" button that copies the bare invite code (e.g. `ABC123`). To actually share an invite, a member has to manually construct the full URL: `https://app.intentionalsociety.org/signup?invite=ABC123`. Most members will just paste the code and wonder why it doesn't work.

## Fix
Change the copy button to copy the full signup URL instead:
```
https://app.intentionalsociety.org/signup?invite=<code>
```

The URL should use `window.location.origin` so it works on preview deployments too.

## File
`src/app/invites/invites-panel.tsx` — the `copy()` function around line 120.

oolu4236:
## Summary
One-line fix: the "Copy" button on the invites panel now writes the full signup URL to the clipboard instead of the bare code.

Before: `ABC123`
After: `https://app.intentionalsociety.org/signup?invite=ABC123`

Uses `window.location.origin` so preview deployments get the correct host automatically.

Closes #202.

### DIFF
--- a/src/app/invites/invites-panel.tsx
+++ b/src/app/invites/invites-panel.tsx
@@ -119,7 +119,8 @@ export function InvitesPanel({ me }: { me: Me }) {
 
   const copy = async (code: string) => {
     try {
-      await navigator.clipboard.writeText(code);
+      const link = `${window.location.origin}/signup?invite=${code}`;
+      await navigator.clipboard.writeText(link);
       setCopiedCode(code);
       setTimeout(() => {
         setCopiedCode((c) => (c === code ? null : c));
--- a/src/app/invites/invites-panel.tsx
+++ b/src/app/invites/invites-panel.tsx
@@ -119,7 +119,7 @@ export function InvitesPanel({ me }: { me: Me }) {
 
   const copy = async (code: string) => {
     try {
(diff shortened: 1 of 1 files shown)


src/app/invites/invites-panel.tsx
```diff
@@ -119,7 +119,8 @@ export function InvitesPanel({ me }: { me: Me }) {
 
   const copy = async (code: string) => {
     try {
-      await navigator.clipboard.writeText(code);
+      const link = `${window.location.origin}/signup?invite=${code}`;
```
james-baker: AI says: #208 — does not work as written. It builds …/signup?invite=${code}, but I checked src/app/signup/page.tsx: the signup
  page reads searchParams: { code?: string } and prefills from code. The copied link should be ?code=${code}. As-is, the
   pasted link lands on signup with an empty code field. Classic missing-context miss — the agent guessed the param
  name. Trivial fix, but it can't merge until that's corrected.

I looked at invites/page.tsx, it does seem to read "code"

james-baker (dismissed):

PR #210, opened 2026-05-17
Show 'Invited by' on member profile pages (#203)

### LINKED ISSUE #203: Member profile page doesn't show who invited them
## Problem
`referredBy` (UUID of the inviting member) is stored in the database and returned in `ProfileForMember`, but `/members/[id]` never displays it. The invite chain is one of the core social signals in IS — knowing who brought someone into the community is meaningful context.

## Goal
Show "Invited by [Name]" on the member profile page below the member-since line, linking to the inviter's profile. If `referredBy` is null (founding members, pre-app members) show nothing.

## Notes
- `ProfileForMember` already includes `referredBy` — just not rendered
- Will need a second lookup or join to get the inviter's display name/slug
- `referredByLegacy` is a text field for pre-app members — could show that as plain text fallback

oolu4236:
## Summary
- Adds `invitedBy` (id, displayName, slug) and `referredByLegacy` to `ProfileForMember` via a self-join
- Member profile page shows: `Member since 2026 · Invited by Aria Chen` with a link to the inviter's profile
- Falls back to `referredByLegacy` (plain text) for pre-app members imported from the Google Sheet
- Shows nothing when `referredBy` is null (founding members, no referrer)

## Test plan
- [ ] Visit `/members/[id]` for a member who was invited — "Invited by [Name]" appears, linking to inviter's profile
- [ ] Visit profile of a founding member — no "Invited by" shown
- [ ] Click inviter link — navigates to their profile

Closes #203.

### DIFF
--- a/src/app/members/[id]/page.tsx
+++ b/src/app/members/[id]/page.tsx
@@ -1,5 +1,6 @@
 import Link from "next/link";
 import { redirect } from "next/navigation";
+import type { UrlObject } from "node:url";
 
 import { Avatar } from "@/components/avatar";
 import { serverApiClient } from "@/lib/api-server";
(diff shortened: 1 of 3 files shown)


james-baker:
Nah, I just plain don't think we should make "who invited you" a permanent part of your identity on this app

PR #245, opened 2026-05-21
Add quarterly intentions field to profiles (#243)

### LINKED ISSUE #243: enhancement of currentIntention field on profiles for slow-cadence dated updates
## Problem
The existing `liveDesire` field is a persistent profile statement. There's no mechanism for members to share what they're currently focused on in a time-stamped, periodically-refreshed way. James raised this in the May 14 meeting as a lower-cadence alternative to a social feed — something like "what I'm working on this quarter" rather than daily status updates.

## Goal
Add a `currentIntention` field to profiles that members can update periodically (roughly quarterly). Show it on member profile pages alongside when it was last updated, so others can see what someone is currently focused on.

## Implementation
- Add `current_intention text` and `intention_updated_at timestamptz` columns to `profiles`
- Expose via `PUT /me` (editable field) and `GET /api/members/:id` (public)
- Show on `/members/[id]` with a "Last updated [month year]" label
- Edit via `/profile/edit` — standard text field

## Distinction from liveDesire
`liveDesire` is a persistent aspiration ("what kind of world/life I want"). `currentIntention` is a present-tense, time-stamped focus ("what I'm actually working on right now").

oolu4236:
Closes #243

## Summary
- Adds `current_intention` and `intention_updated_at` columns to `profiles` (migration `0011_add_quarterly_intention.sql`)
- Exposes both fields via `PUT /me` (added to `EDITABLE_PROFILE_FIELDS`)
- Displays `currentIntention` with a formatted date on member profile pages
- Adds a "Current intention" textarea to the edit profile form
- Updates functional tests to include the new fields in shape assertions

## Test plan
- [ ] Run `npm run dev` — the migration applies automatically via the `dev:db` step
- [ ] Edit your profile and set a current intention — verify it saves and appears on your public profile
- [ ] Visit another member's profile — confirm their intention and date show when set, and field is absent when blank
- [ ] Confirm `emergencyContact` and `isAdmin` are still absent from the member profile view


james-baker:
@oolu4236 I notice "Run npm run dev:db:reset locally to apply the migration" and that's incorrect: "npm run dev" includes "dev:db" which includes the migration script, so you do not need to reset your database in order to apply migrations.

drizzle/0011_add_quarterly_intention.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "profiles" ADD COLUMN "current_intention" text;--> statement-breakpoint
+ALTER TABLE "profiles" ADD COLUMN "intention_updated_at" timestamp with time zone;
```
james-baker: I believe a data-migration step will show up here when added.

src/components/profile-fields.tsx
james-baker: Replacing this?

james-baker (changes requested):
I'm sorry, I thought I reviewed this already but I'm not seeing anything here now.

I don't want to have this next to liveDesire - I want to replace liveDesire with currentIntention (so, it's a rename with some polishing). They're the same thing.

drizzle/0011_add_quarterly_intention.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "profiles" ADD COLUMN "current_intention" text;--> statement-breakpoint
+ALTER TABLE "profiles" ADD COLUMN "intention_updated_at" timestamp with time zone;--> statement-breakpoint
+UPDATE "profiles" SET "current_intention" = "live_desire", "intention_updated_at" = now() WHERE "live_desire" IS NOT NULL AND "current_intention" IS NULL;
```
james-baker: use profiles.created_at instead of "now" - I deliberately set that with the date people filled out the Google Form

drizzle/meta/0007_snapshot.json
```diff
@@ -345,13 +345,6 @@
           "notNull": true,
           "default": false
         },
-        "hidden": {
```
james-baker, line 348: Huh, what happened here? I don't think the "hidden" column should go away.... OHH this should not be 0007, this should be 11 right? maybe this file needs to be reverted?

james-baker (changes requested):
the rest looks good, but let's get the 0007 cruft out (and might as well improve the created_at time since we're doing that)

james-baker (approved):
LGTM!

james-baker:
schema has been deployed a few hours ago

james-baker:
Found and fixed "out of order timestamps on migrations" problem.

james-baker:
@oolu4236 the old column can be removed at your leisure.

PR #246, opened 2026-05-21
Re-center web graph on node click; navigate on double-click (#239)

### LINKED ISSUE #239: Re-center web graph on node click instead of navigating away
## Problem
Clicking a node on the My Web graph (`/myweb`) immediately navigates to that member's profile page via `router.push()`. This breaks the flow — you lose your position in the graph and have to navigate back to continue exploring.

## Goal
Single-click on a node should re-center the graph view on that node (using ReactFlow's `fitView` or `setCenter` on that node's position). Double-click (or a "View profile" button that appears on the selected node) should navigate to the member's profile.

## Current code
`src/app/myweb/web-graph.tsx` line 311 — `onNodeClick` calls `router.push()` directly.

## Notes
- Center node (self) is already a no-op on click — that behaviour should stay
- ReactFlow exposes `useReactFlow().setCenter(x, y)` for programmatic panning
- The node's `position` from the force simulation gives the coordinates to center on

oolu4236:
Closes #239

## Summary
- Single-click on a graph node now re-centers the canvas on that node (animated, 400 ms) instead of navigating away
- Double-click navigates to the member's profile page
- The center node (self) is still a no-op for both interactions

## Test plan
- [ ] Open `/myweb` and single-click a connected member node — canvas should smoothly pan/zoom to center on it
- [ ] Double-click the same node — should navigate to `/members/[slug]`
- [ ] Clicking the center (self) node should do nothing

### DIFF
--- a/src/app/myweb/web-graph.tsx
+++ b/src/app/myweb/web-graph.tsx
@@ -312,8 +312,16 @@ export function WebGraph({ onOpenRelating }: { onOpenRelating: (target: Relating
         elementsSelectable={false}
         proOptions={{ hideAttribution: true }}
         onNodeClick={(_event, node) => {
-          // Center node has no profile destination distinct from "you
-          // are here" — clicking yourself is a no-op.
+          if (node.data.isCenter) return;
+          // Re-center the viewport on the clicked node so members can
+          // explore the graph without navigating away. Double-click
+          // navigates to the member's profile page.
+          flowRef.current?.setCenter(node.position.x, node.position.y, {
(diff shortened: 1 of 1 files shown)


james-baker:
Feels like this could have been bundled with the other (one-liner) change to the web.

I thought the real issue here was that there was no hand-pointer cursor when over the nodes, so that you didn't know that a click would be a "link click". I don't know if double-click is going to be discoverable.

But let's try it and see I guess!

PR #247, opened 2026-05-21
Add mailto contact link on member profile pages (#242)

### LINKED ISSUE #242: Show mailto: contact link on member profile pages
## Problem
Members have no way to contact each other through the app. The May 14 team meeting reached consensus that the right approach is surfacing email addresses with a mailto: link rather than building in-app messaging.

## Goal
Show a "Send email" link (mailto:) on `/members/[id]` so members can initiate contact directly through their email client.

## Implementation
- The member's email comes from `auth.users`, not `profiles` — requires a join or separate lookup in `getProfileForMember`
- Only show to authenticated members (already the case — the page requires auth)
- Style as a subtle secondary action below the profile fields, not a prominent CTA

## Privacy note
Email is only visible to signed-in IS members, not the public. This matches the spirit of a private community app.

oolu4236:
Closes #242

## Summary
- Adds a "Send email" link (`mailto:`) to member profile pages when the member has an email address on file
- Email is fetched via a join on `auth.users` in `getProfileForMember` — only visible to authenticated members (the existing auth guard on the page)
- `emergencyContact` remains self/admin-only; email is the contact affordance for peers

## Test plan
- [ ] Visit another member's profile — a "Send email" link should appear and open the system mail client
- [ ] Verify the link is absent if somehow no email is present
- [ ] Confirm `emergencyContact` is still NOT visible on another member's profile

### DIFF
--- a/src/app/members/[id]/page.tsx
+++ b/src/app/members/[id]/page.tsx
@@ -64,6 +64,15 @@ export default async function MemberProfilePage({ params }: { params: Promise<{ 
         <Field label="Live desire">{profile.liveDesire}</Field>
         <Field label="Supplementary info">{profile.supplementaryInfo}</Field>
       </dl>
+
+      {profile.email && (
+        <a
+          href={`mailto:${profile.email}`}
+          className="text-base text-muted-foreground underline hover:text-foreground"
+        >
+          Send email
+        </a>
+      )}
     </main>
   );
 }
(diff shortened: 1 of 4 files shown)


src/app/members/[id]/page.tsx
```diff
@@ -64,6 +64,15 @@ export default async function MemberProfilePage({ params }: { params: Promise<{
         <Field label="Live desire">{profile.liveDesire}</Field>
         <Field label="Supplementary info">{profile.supplementaryInfo}</Field>
       </dl>
+
+      {profile.email && (
+        <a
+          href={`mailto:${profile.email}`}
+          className="text-base text-muted-foreground underline hover:text-foreground"
+        >
+          Send email
```
Comment (james-baker) [inline src/app/members/[id]/page.tsx:73]: I'd rather the display text be the email address, but this is still an improvement so, approving

james-baker (approved):

PR #249, opened 2026-05-21
Add hover tooltip to web graph nodes (#240)

### LINKED ISSUE #240: Show member name tooltip on web graph node hover
## Problem
Nodes on the My Web graph show an avatar and a small display name label, but when the graph has many nodes the labels overlap or get clipped. There's no hover state that clearly identifies who a node is.

## Goal
Show a tooltip with the member's display name when hovering over a node. Should appear near the node without obscuring it, and disappear when the cursor moves away.

## Implementation notes
- ReactFlow nodes support `title` on their wrapper div for a native browser tooltip (quick win, no extra component)
- Or use a small styled overlay positioned relative to the hovered node for better visual control
- Tooltip should show `displayName` — already available on `MemberNodeData`
- Should not appear on the center node (self)

oolu4236:
Closes #240

## Summary
- Adds a native `title` attribute to each member node in the web graph showing the member's display name on hover
- No tooltip on the center (self) node — the name is already visible there

## Test plan
- [ ] Open `/myweb` and hover over a connected member node — a tooltip with their name should appear
- [ ] Confirm the center node has no tooltip

### DIFF
--- a/src/app/myweb/web-graph.tsx
+++ b/src/app/myweb/web-graph.tsx
@@ -76,7 +76,10 @@ const HANDLE_STYLE = { opacity: 0, top: "50%", pointerEvents: "none" as const };
 
 function MemberNode({ data }: NodeProps<Node<MemberNodeData>>) {
   return (
-    <div className="flex flex-col items-center gap-1">
+    <div
+      className="flex flex-col items-center gap-1"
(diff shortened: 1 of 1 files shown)


james-baker:
LGTM

PR #253, opened 2026-05-21
Surface PR link and migration SQL on expand-schema approval gate

james-baker:
## Summary

Splits `forward-migrate-prod-schema-expansion.yml` into two jobs so the prod-db reviewer has context *before* approving:

- **`prep`** (ungated) — checks out the dispatched ref, looks up the open PR via `gh pr list --head`, runs the migration preview, and runs the destructive-migration guard. Writes a markdown summary (PR link + changed files + full SQL of every new migration) to `$GITHUB_STEP_SUMMARY`. Surfaces `pr_url` as a job output.
- **`migrate`** (`needs: prep`, gated on `prod-db`) — consumes `needs.prep.outputs.pr_url` as `environment.url`, runs `drizzle-kit migrate`.

Also adds a top-level `run-name: "Expand prod schema for ${{ inputs.ref }}"` so the Actions list shows the dispatched ref directly.

## What changes for the reviewer

- Actions list title: branch/SHA is visible without clicking in.
- Approval card: "View deployment" link → opens the PR the migration is coming from. Falls back to the repo root if the ref has no open PR (e.g., a direct SHA).
- Workflow run page (before approval): markdown summary with the migration SQL.
- A destructive migration (`DROP COLUMN`, `RENAME`, etc.) fails `prep`, so reviewers no longer get pinged for a run that would self-reject post-approval.

## Tested

The `gh pr list --head $ref` lookup verified against three cases:
- Open PR (`hidden-accounts`) → returns the PR URL.
- Branch with no open PR (`main`) → empty stdout, exit 0 → fallback fires.
- A SHA (realistic non-branch dispatch) → empty stdout, exit 0 → fallback fires.

No 3rd-party actions added — `actions/checkout@v6` and `actions/setup-node@v6` are the same first-party actions the workflow already used. `gh` is preinstalled on the runner.

## Permissions

Adds `pull-requests: read` so `gh pr list` works. The existing `contents: read` is unchanged.

### DIFF
--- a/.github/workflows/forward-migrate-prod-schema-expansion.yml
+++ b/.github/workflows/forward-migrate-prod-schema-expansion.yml
@@ -1,4 +1,5 @@
 name: Forward-migrate prod schema (expansion)
+run-name: "Expand prod schema for ${{ inputs.ref }}"
 
 # Manually triggered. Applies additive (expand) drizzle migrations to
 # the prod DB ahead of opening a PR, so previews can exercise new code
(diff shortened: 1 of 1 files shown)


james-baker:
I AM NOT COMPROMISED and I want to test this immediately - bypassing.

PR #277, opened 2026-05-25
feat: admin member toggle with admin badge in nav

### LINKED ISSUE #251: Admin Page: toggle admin status for users
Admins will be able to change the admin status of users. 

The last admin will not be able to toggle themselves off.

AlexisChen99:
Closes #251

## Summary

- Adds a **Members** section to `/admin` with a toggle button per member to grant or revoke admin status
- **Self-demotion guard**: your own toggle is disabled in the UI; the API returns 403 if attempted
- **Last-admin guard**: the API returns 409 if removing the last remaining admin
- Members sorted alphabetically by display name with a stable `id` tiebreaker
- Adds a `ShieldCheck` icon to the Admin nav link and a light-green **Admin** badge in the sheet header, visible only to admins

## Implementation notes

- `src/server/members-admin.ts` — `listAdminMembers()` + `setAdminStatus()` with both guards
- `GET /admin/members` + `PATCH /admin/members/:id/admin` added to the admin sub-router
- UUID validation in the route handler (returns 400), consistent with adjacent routes
- Panel uses `useQuery` + `useMutation` (same pattern as `AdminHidden`), with `staleTime: 5min` to avoid spurious refetches on tab focus
- Known limitation: last-admin count check and update are separate statements; a concurrent double-demotion could theoretically bypass the guard. Transactions over the Supabase pooler are unreliable (see `docs/strategy-db-transactions.md`), so a comment documents the trade-off.

## Test plan

- [ ] As admin: open `/admin`, scroll to Members section — list appears sorted A→Z
- [ ] Toggle a non-admin member to admin — button changes to "Remove admin", refetch shows updated state
- [ ] Toggle an admin member to non-admin — works when ≥2 admins exist
- [ ] Try to remove the last admin — error message "Cannot remove the last admin."
- [ ] Your own row shows "Admin (you)" (disabled) — cannot toggle self
- [ ] Open the nav menu as admin — ShieldCheck icon on Admin link + green Admin badge in header
- [ ] Open the nav menu as non-admin — no badge, no Admin link

### DIFF
--- a/src/server/api.ts
+++ b/src/server/api.ts
@@ -56,7 +56,7 @@ import {
   parseOptionalRelationValue,
   updateRelationValue,
 } from "./relations";
-import { listAdminMembers, setAdminStatus } from "./members-admin";
+import { listMembersAdmin, setAdminStatus } from "./members-admin";
 import { profiles } from "./schema";
 import { resetE2EUsers } from "./test-reset";
 
@@ -214,7 +214,7 @@ const adminRoutes = new Hono<{ Variables: ApiVariables }>()
     return c.json(result);
   })
   .get("/members", async (c) => {
-    const members = await listAdminMembers();
+    const members = await listMembersAdmin();
     return c.json({ members });
   })
   .patch(
--- a/src/server/api.ts
+++ b/src/server/api.ts
@@ -56,6 +56,7 @@ import {
   parseOptionalRelationValue,
   updateRelationValue,
 } from "./relations";
+import { listAdminMembers, setAdminStatus } from "./members-admin";
 import { profiles } from "./schema";
 import { resetE2EUsers } from "./test-reset";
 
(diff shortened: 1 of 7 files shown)


src/server/api.ts
```diff
@@ -170,6 +171,35 @@ const adminRoutes = new Hono<{ Variables: ApiVariables }>()
       if ("error" in result) return c.json({ error: result.error }, 404);
       return c.json({ ok: true });
     },
+  )
+  .get("/members", async (c) => {
```
james-baker, line 216: is this really "list-admins" ?

james-baker: actually I was wrong, it's not! it lists all members, and the "listAdminMembers" function name confused me, so Alexis was right to change it.

Why is there both @AlexisChen99 and @alexisChen9090 here, what's going on are there two accounts?

james-baker:
resolving merge conflicts and merging...

PR #282, opened 2026-05-26
feat: redesign sign-in form with explicit email/password mode toggle

### LINKED ISSUE #268: [URGENT] Sign-in: optional password field is confusing — improve copy and UX
## Background

User testing revealed confusion around the optional password field in the sign-in and onboarding flows. Users weren't sure whether they *need* to create a password, what happens if they don't, and what the relationship is between password sign-in and email sign-in.

## Observed problems

- Users tried to create a password when they didn't need to, or didn't realise they could skip it.
- The optionality of the password is not communicated clearly enough — \"optional\" in a label isn't enough when users don't yet understand the two-mode sign-in model.
- The conceptual split between \"sign in by email link\" and \"sign in by password\" isn't explained inline.

## Related issues

- #225 — replaces \"magic link\" jargon with plain language (addresses some of the wording)
- #212 — remove-password option (complement to this, once password is set)
- #211 — unhandled error in password-update step of welcome form

## Suggested fixes

1. Add a short inline explanation (e.g. a tooltip or helper paragraph) explaining that email sign-in always works and a password is only an optional convenience.
2. Make the skip/leave-blank path visually obvious — consider a dedicated \"Skip, I'll use email sign-in\" affordance rather than just a greyed-out placeholder.
3. Review the welcome-form step order: does password setup come too early before users understand the product?
4. Coordinate copy changes with #225 so wording is consistent after both land.

## Priority

Surfaced in first user test. Fix before next user test session.

AlexisChen99:
## Summary

- Replace the always-visible optional password field with a two-mode toggle: email link (default) and password (user-initiated)
- Password field only renders when in password mode — keeps autofill / password managers working correctly
- "Forgot your password?" link moves from the page footer into the form, contextually below the password field
- Primary CTA uses `variant="secondary"` (filled teal); mode switch uses `variant="ghost"` for clear visual hierarchy
- Page subtitle simplified to "Enter your email and we'll send you a sign-in link." to match default mode

## Test plan

- [ ] Default state: only email field + "Send sign-in link" (teal) + "Sign in with password instead" (ghost) visible
- [ ] Click "Sign in with password instead": password field appears, focus moves to it, button becomes "Sign in"
- [ ] "Forgot your password?" link appears below password field only in password mode
- [ ] Click "Use email link instead": reverts to email mode, password field hidden, error cleared
- [ ] Password autofill works (field not in DOM until needed)
- [ ] Submitting with email only sends OTP link; submitting with password signs in directly

Closes #268

### DIFF
--- a/src/app/signin/page.tsx
+++ b/src/app/signin/page.tsx
@@ -36,7 +36,7 @@ export default async function SigninPage({ searchParams }: SigninPageProps) {
     <main className="flex min-h-screen flex-col items-center justify-center gap-6 p-8">
       <AppWordmark asLink />
       <p className="max-w-sm text-center text-base text-muted-foreground">
-        Receive a sign-in link by email, or sign in with your password if you've set one.
+        Enter your email and we'll send you a sign-in link.
       </p>
       {process.env.NODE_ENV === "development" && (
         <p className="max-w-sm rounded border border-amber-500/40 bg-amber-500/10 px-4 py-3 text-sm text-amber-700 dark:text-amber-400">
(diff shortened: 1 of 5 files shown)


AlexisChen99:
Screenshot 2026-05-25 at 6 43 14 PM
Screenshot 2026-05-25 at 6 43 17 PM

james-baker:
Yesss, love it.

james-baker:
But you forgot to run the e2e tests (or "npm test"). Lot of tests rely on sign in UI. :)  I'll fix if I get to it first.

PR #301, opened 2026-05-27
Add member account deactivation

oolu4236:
Closes part of #185.

**deactivate** is self-serve for members; full **delete** remains an admin-only operation.

## What this adds

- `deactivated_at` timestamp column on `profiles` (migration `0011`)
- `POST /me/deactivate` API endpoint — sets `deactivated_at` to now
- Confirmation UI on the Edit Profile page (two-step: button → confirm/cancel)
- Deactivated profiles will be hidden from all member-facing views (directory, typeaheads, member page) — filter enforcement to follow in the contract PR once the column is live

## What's deferred

- Filtering deactivated profiles out of member listings (expand-contract: column lands first)
- Admin reactivation UI (admin can clear `deactivated_at` directly for now)
- Full account delete remains out of scope per #185

## Test plan
- [ ] Run `npm run dev` — migration applies cleanly
- [ ] Log in as a test member, go to Edit Profile, see "Deactivate account" section
- [ ] Click Deactivate → confirm step appears
- [ ] Confirm → redirected to `/signin`
- [ ] Verify `deactivated_at` is set in DB


src/app/profile/edit/page.tsx
```diff
@@ -45,6 +46,14 @@ export default async function EditProfilePage() {
         </p>
         <ChangePasswordForm />
       </div>
+
+      <div className="flex w-full max-w-md flex-col gap-3 border-t border-border pt-6">
+        <h2 className="text-base font-semibold">Deactivate account</h2>
+        <p className="text-sm text-muted-foreground">
+          Hides your profile from other members. An admin can reactivate it for you at any time.
```
james-baker: need something like "Some records of past participation may remain visible in other people's history data."

Can they still sign in? It looks like nothing blocks them.

If so, why do they need an admin to reactivate?

src/app/profile/deactivate-account-form.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <Button variant="destructive" onClick={() => setConfirming(true)}>
+        Deactivate account
+      </Button>
+    );
+  }
+
+  return (
+    <div className="flex flex-col gap-3">
+      <p className="text-sm text-muted-foreground">
+        This will hide your profile from all other members. You can ask an admin to reactivate your account at any
```
james-baker: why is this text duplicated inside and outside the form?

james-baker (changes requested):
Where are the effects? (e.g. other people don't see them appear in the directory, maps, hints)

The migration will need to be re-stamped.

Also I'd like some tests added before this goes through.

oolu4236:
@james-baker  I've tested updates locally and everything is working as expected.

When a member is deactivated, they disappear from the directory and web graph, but can still sign in and reactivate their account from the edit page.

I’ve also incorporated the feedback: the necessary areas are connected (directory, member lookups, relation suggestions, and web graph), members can reactivate themselves without admin help, duplicate copy has been removed, a participation note has been added, the migration has been re-stamped, and functional tests are included.

The only outstanding issue is E2E testing, which is failing because the `deactivated_at` column hasn’t been added to production yet. The `prod:db:expand `step is currently waiting for approval.

docs/devjournal.md
```diff
@@ -4,6 +4,18 @@ Each entry: **Date** | **Author** | **Title**, followed by description text. Mos
 
 ---
 
+## 2026-05-30 | Ola | Deactivate account (member self-serve, #185)
+
+Members can deactivate their own account from the edit-profile page (`POST /me/deactivate`). Deactivated profiles are filtered from the directory, member lookups, relation suggestions, and the web graph — same `IS NULL` guard pattern as `profiles.hidden`. Members can reactivate themselves at any time via the same page (`POST /me/reactivate`); no admin action needed. `deactivated_at` is exposed in `GET /me` so the UI can show state. Functional tests cover deactivate, reactivate, and the self-shape assertion. (#301)
+
+## 2026-05-30 | Ola | Drop live_desire column (#319)
```
james-baker: this is a really good example of something that doesn't need a journal entry

drizzle/meta/_journal.json
```diff
@@ -85,6 +85,13 @@
       "when": 1780079305136,
       "tag": "0011_add_quarterly_intention",
       "breakpoints": true
+    },
+    {
+      "idx": 12,
```
james-baker: both this PR and the "drop the column" PR made this index 12, so this one now gets bumped to 13

src/server/relations.ts
```diff
@@ -321,7 +323,9 @@ export const getPersonalWeb = async (params: {
     const rows = await db
       .select({ id: profiles.id })
       .from(profiles)
-      .where(and(inArray(profiles.id, [...candidateIds]), eq(profiles.hidden, true)));
+      .where(
+        and(inArray(profiles.id, [...candidateIds]), or(eq(profiles.hidden, true), isNotNull(profiles.deactivatedAt))),
```
james-baker: could you add one functional test on relations that verifies that a deactivated profile, with a relationship to the test user, still doesn't include the deactivated user in their map-data?

james-baker (commented):
One test and one migration restamp, looks a lot better.

I trust that you've verified all the behavior once manually?

And one question: if you're using the "suggest a member" from the "create an invite" screen, does that also filter out the deactivated?

docs/devjournal.md
james-baker: (this comment is on line 11, the "drop an unused column and nothing changes" entry)

oolu4236:
> One test and one migration restamp, looks a lot better.
> 
> I trust that you've verified all the behavior once manually?
> 
> And one question: if you're using the "suggest a member" from the "create an invite" screen, does that also filter out the deactivated?

Tested locally and everything is working as expected. When a member is deactivated, they are hidden from the directory, web graph, and relation suggestions. They can still sign in and reactivate their account from the edit page. 

The invite suggestions case is already covered. `MemberTypeahead` uses `GET /members`, which goes through `listMembers`, and that query already filters out deactivated members with `deactivated_at IS NULL`.

docs/devjournal.md
oolu4236: > this is a really good example of something that doesn't need a journal entry

    confirming that deactivated members remain hidden even when includeHidden=true (admin view).

drizzle/meta/_journal.json
oolu4236: > both this PR and the "drop the column" PR made this index 12, so this one now gets bumped to 13

I've renumbered the migration to 0013, since drop-live-desire now uses 0012.

src/server/relations.ts
oolu4236: > could you add one functional test on relations that verifies that a deactivated profile, with a relationship to the test user, still doesn't include the deactivated user in their map-data?

Added one verifying that deactivated members are hidden for regular users and confirming that deactivated members remain hidden even when includeHidden=true (admin view).

james-baker:
resolving conflicts

james-baker:
one fix happening inside the merge:

 While validating, the suite surfaced a pre-existing bug in the branch (not caused by the merge — the branch tip fails this
  test too):

  relations.ts gated deactivated-profile filtering behind if (!includeHidden), so an admin viewing the web with
  includeHidden=true would still see deactivated members — but the branch's own test, "includeHidden=true still drops
  deactivated neighbors", asserts they shouldn't (deactivation ≠ hidden; only hidden is admin-bypassable). I introduced a
  single dropFilter reused in both hops: deactivated is always dropped; hidden only for non-admins. That matches the test and
  the devjournal's "filtered from … the web graph."

james-baker (approved):
now cleared for landing

PR #304, opened 2026-05-27
feat: portable AI procedure framework: /commit /pr /ship (#133)

NorsemanSpiff:
## Summary

Implement the three portable AI procedure Skills (`/commit`, `/pr`,
`/ship`) under `.claude/skills/` per the merged v5 spec (PR #133,
`docs/spec-portable-ai-procedures.md`) and the merged implementation
plan (PR #283, `docs/plan-portable-ai-procedures.md`). This PR was
itself partially self-hosted -- its two commits were produced via the
new `/commit` Skill's bundled-approval flow -- and, on approval, the
final merge will run via `/ship` on this PR itself.

## Why

The team's check-in workflow has so far depended on each teammate's
personal prompting and memory, so execution drifts across teammates
and across tool switches. Encoding the workflow as explicit Claude
Code Skills lets any teammate invoke `/commit`, `/pr`, `/ship` and
get the same behavior: local `npm test` gate, deterministic staging,
bundled human approval, merge-commit-only shipping, post-merge watch
on `main`.

This PR carries the implementation payload of dev-plan Phases 7a, 7b,
and 7c. Phase 8 validation (per-Skill eval self-review + the end-to-
end smoke-test on PR #303, merged 2026-05-27 at `7fbee8e`) is
complete. The Phase 9 self-host -- running `/ship` on this PR to merge
it -- is the integration test that earns the Skills their reputation
by shipping themselves; that step waits for your review approval.

## Behavior

Two commits on this branch:

**1. `feat: add /commit /pr /ship Skills + eval definitions` (Phase 7a + the `.gitignore` adjustment 7c needed)**

- `.claude/skills/commit/SKILL.md` (183 lines, under the spec's 500-line cap) -- Conventional Commit subject draft, suspicious-file blocker, combined expand+contract refusal, single bundled human approval, push.
- `.claude/skills/pr/SKILL.md` (93 lines) -- fetch + freshness rebase + re-run `npm test` if rebase changed anything, push, draft CC-format PR title/body on fresh branches, comment-by-default on existing PRs, refuses branch-switching.
[...]

Issue #285, opened 2026-05-26
fix: add npx playwright install to setup docs and npm run setup


NorsemanSpiff:
## Problem Statement

Running `npm test` or `npm run test:e2e` on a machine where Playwright browsers have not been installed fails for all e2e tests with:

```
Error: browserType.launch: Executable doesn't exist at
C:\Users\...\ms-playwright\chromium_headless_shell-1223\chrome-headless-shell-win64\chrome-headless-shell.exe

╔════════════════════════════════════════════════════════════════════╗
║ Looks like Playwright was just installed or updated.               ║
║ Please run the following command to download new browsers:         ║
║     npx playwright install                                         ║
╚════════════════════════════════════════════════════════════════════╝
```

The Playwright browser binaries are not downloaded by `npm install` — they require a separate `npx playwright install` step. This step is not currently documented in `docs/setup-dev-machine.md` and is not automated by `npm run setup`.

**Observed in RCA #284 (2026-05-25):** All 25 e2e tests failed with this error on a machine that had run `npm install` but not `npx playwright install`. Functional tests (Vitest, 272/272) passed normally. CI was unaffected (it installs browsers via its own workflow step).

## Scope

- `docs/setup-dev-machine.md` — add `npx playwright install` as an explicit, required setup step
- `scripts/setup.mjs` — optionally automate the step so `npm run setup` handles it

## Proposed Implementation Approach

**Option A — Docs only (minimum, immediate):**

In `docs/setup-dev-machine.md`, add after the `npm install` step:

```
npx playwright install   # downloads Chromium browser binary for e2e tests
                         # safe to re-run; only downloads if missing or outdated
```

**Option B — Automate in `npm run setup` (recommended):**

Add a step to `scripts/setup.mjs`:

```js
import { execSync } from "node:child_process";

function ensurePlaywrightBrowsers() {
  console.log("2. Playwright browser binaries");
  try {
    execSync("npx playwright install chromium", { stdio: "inherit" });
  } catch {
    console.error("  Warning: playwright install failed — e2e tests will not run.");
    console.error("  Try manually: npx playwright install");
  }
}
```

Installing only `chromium` (rather than all browsers) is sufficient — the e2e suite uses Chromium only (see `playwright.config.ts`). The install is idempotent: Playwright skips download if the binary is already current.

**Note:** `npx playwright install` downloads ~100MB of browser binaries per browser. Scoping to `chromium` only keeps the install lean.

## Acceptance Criteria

[...]

PR #307, opened 2026-05-28
feat: wire Biome auto-format at 3 layers + baseline format pass (#140)

### LINKED ISSUE #140: Consider adding automatic (post-edit) formatting to dev workflow
I've installed an auto-formatter via Biome that's available via "npm format" but not wired into anything automatic yet.

Desired: Consider whether any auto-formatting is useful, and if so where should we wire it in? The answer could be "nah, we can just run it when we feel like it", or one or both of the below, or something I haven't even thought of.

I asked Claude about two options I know of (I don't know if there are more things I should have asked):

## Git pre-commit vs Claude post-edit hooks

These are different tools for different blast radii. Both can run Biome.

### Git pre-commit hook
- **Where:** `.git/hooks/pre-commit` (private to each clone) — or made shareable via a tool like **husky**, **simple-git-hooks**, or **lefthook** that wires `.git/hooks` to a version-controlled config in the repo.
- **When it fires:** before `git commit` finalizes. Either runs every check (slow) or — with **lint-staged** — runs only on the files actually staged.
- **What you'd put in it:** `biome check --write` on staged files, then re-stage them. So formatting is fixed on the way in; the commit itself is always clean.
- **Coverage:** *everyone* committing to this repo, with any tool — me, you in your editor, a teammate using vim, an agent that bypasses Claude. Fully team-wide.
- **Bypass:** `git commit --no-verify` skips it.
- **Cost:** small dependency (one of those tools above), ~50 ms wakeup before each commit.

### Claude post-edit hook
- **Where:** `.claude/settings.json` (project-scoped) or `~/.claude/settings.json` (user-scoped).
- **When it fires:** after I use `Edit` or `Write` on a file. The hook receives JSON on stdin with the path I just touched.
- **What you'd put in it:** `biome format --write` (or `biome check --write`) on just that file path. So my edits are reformatted before you see them in the conversation, on the order of milliseconds.
- **Coverage:** *only my edits.* Manual edits, vim edits, edits from another agent — not covered.
- **Bypass:** none from my side. You can disable in settings.

### Yes, "post-file-edit" is exactly the Claude hook event
Claude Code's `PostToolUse` event with matcher `Edit|Write` is precisely what you're describing. The hook runs after each file mutation and can do anything a shell command can. It's what enforces "Claude never hands me unformatted code."

### Which one to pick

They're complementary:

|                                          | Git pre-commit                   | Claude post-edit                      |
| ---------------------------------------- | -------------------------------- | ------------------------------------- |
| Catches my unformatted edits             | yes (eventually, at commit time) | yes (immediately)                     |
| Catches your manual edits                | yes                              | no                                    |
| Catches a teammate's edits               | yes                              | no                                    |
| Catches an agent that doesn't use Claude | yes                              | no                                    |
| Visible feedback in conversation         | no                               | yes (you see clean code in real-time) |

**If you pick one, it's the git pre-commit.** It's universal coverage, the standard pattern, and works regardless of editor or agent. The Claude hook is a nice quality-of-life on top — it makes my output cleaner without you needing to commit-and-fix.

**If you want my recommendation:** add the git pre-commit hook (via `lefthook` — fastest to set up, single binary, no Node deps) for team-wide coverage. Skip the Claude hook unless you start noticing me handing you unformatted snippets often.

Want me to wire up lefthook in this branch, or punt that to a follow-up?

NorsemanSpiff:
Closes #140. Discussion: https://github.com/Intentional-Society/is-app/issues/140#issuecomment-4561265025.

## Summary

Wires Biome enforcement into three layers so format violations are caught automatically rather than at PR review time:

1. **`npm test`** — `lint` script changed from `biome lint .` to `biome check .`. Now runs lint + format + organize-imports as a single read-only gate.
2. **Pre-commit (lefthook)** — `lefthook.yml` runs `biome check --write` on staged JS/TS/MJS/CJS/JSON/CSS files and re-stages the fixes. Universal coverage (any editor, any agent). Bypass with `--no-verify`. `lefthook install` is wired into `npm run setup` (defense-in-depth — lefthook's own postinstall already handles it for npm-based setups).
3. **CI** — `.github/workflows/ci.yml` swaps `npm run lint` for `npx biome ci .` (same checks, plus GitHub annotations on the PR).

A **per-edit Claude Code hook** was prototyped in the original PR but dropped during review — the ~1 s npx-overhead per AI Edit/Write was deemed too costly. We replaced it with **pre-commit telemetry** instead: `scripts/biome-precommit.mjs` wraps biome and appends one JSONL line to `.scratch/biome-precommit-stats.log` per pre-commit, recording how often a commit needed reformatting. If the data shows reformatting is common, we'll revisit the per-edit hook (with a direct-binary invocation to avoid npx overhead).

Also added (alongside the wiring):
- `.git-blame-ignore-revs` listing the baseline format pass commit (`9a6cd6b`) so `git blame` skips past the formatting noise. `scripts/setup.mjs` runs `git config blame.ignoreRevsFile .git-blame-ignore-revs` so every dev gets the local-git side automatically (GitHub honors the file by default).
- `biome.json` exclusions for `drizzle/meta` and `tests/functional/server/__data__` (both tool-regenerated — biome would otherwise fight drizzle-kit and the buttondown gold recorder).
- 7 vitest tests for the telemetry wrapper's pure helpers.

## Commits

| SHA | Subject |
|---|---|
| `9a6cd6b` | `chore: tighten biome scope + baseline format pass` — auto-fixes ~50 files of accumulated format drift across `src/`, `tests/`, `scripts/`, root configs. Adds two `biome.json` exclusions. Pure whitespace/wrapping/import-organization. Listed in `.git-blame-ignore-revs`. |
| `eb2e0f1` | `feat: wire Biome auto-format into npm test, Claude, pre-commit, and CI` — original 4-layer wiring (Claude hook included). |
| `98d2fd5` | `fix: drop Claude post-edit hook, add pre-commit telemetry + blame config (#307 review)` — addresses all 5 of @james-baker's review comments. Removes the Claude hook (and its `.gitignore` negations + settings.json + hooks/ dir). Adds the telemetry wrapper. Adds the `git config blame.ignoreRevsFile` setup step. |
| `dabab1a` | `test: add vitest coverage for biome-precommit wrapper + widen lefthook glob` — 7 unit tests for the wrapper's pure helpers; extends lefthook glob to `*.{js,jsx,mjs,cjs,ts,tsx,json,jsonc,css}` so `.mjs`/`.cjs` files get formatted at commit time (gap surfaced by shipping #98d2fd5 itself). |

Why both `9a6cd6b` and `eb2e0f1` are in this PR: the baseline has to land before (or with) the `npm test` change, otherwise `npm test` is red on `main`. Splitting across PRs invites the "we'll do the baseline later" trap.

## Reviewer asks still open

Two scope questions worth a second opinion before merge:

1. **Are these the right Biome exclusions?** Added `!drizzle/meta` (drizzle-kit regenerates) and `!tests/functional/server/__data__` (buttondown gold recorder regenerates). Anything else generated/recorded that I missed?
2. **Denylist vs allowlist on `biome.json` `includes`?** Currently denylist (`["**", "!a", "!b"]`). Allowlist (`["src/**", "tests/**", "scripts/**", ...]`) would be more conservative — generated stuff doesn't sneak in by default — but requires explicit additions when new top-level directories appear. Happy to flip if you want.

## Heads-up

- **The diff is mostly the baseline.** ~50 reformatted files in `9a6cd6b` are pure whitespace; the actual wiring is in the other three commits.
- **Merge conflict with PR #304:** that PR also touches the `.claude` block in `.gitignore`. Whichever lands second gets a trivial conflict — combine the relevant lines.
- **`.git-blame-ignore-revs`:** new file at repo root pinning `9a6cd6b`. GitHub honors it automatically. `npm run setup` configures local git too (idempotent).

## Test plan (verified locally)

- `npm run lint` (`biome check .`) → 181 files, 0 errors
- `npm run format` → 0 fixes needed
- `npx biome ci .` → passes
- `npx tsc --noEmit` → passes
- `npm run test:functional` → 22 files / 338 tests pass (includes 7 new tests for the precommit wrapper)
- `npx lefthook install` → installs pre-commit hook
- Lefthook smoke (twice — once with `.ts`, once with `.mjs` to verify glob coverage): mangled-spacing file staged + committed → wrapper invoked, biome reformatted, `stage_fixed: true` re-staged, commit captured the formatted version, stats log entry written.

## Pending verification (post-push)

- CI green on this PR.
- Vercel preview deploy succeeds (no DB changes; deploy risk low).
- Telemetry data from real commits accrues over the next few weeks — answers whether we should reintroduce the per-edit Claude hook.


.claude/hooks/format-file.mjs
```diff
@@ -0,0 +1,35 @@
...
+  if (!raw.trim()) process.exit(0);
+
+  const payload = JSON.parse(raw);
+  const filePath = payload?.tool_input?.file_path;
+  if (typeof filePath !== "string" || !filePath) process.exit(0);
+
+  spawnSync("npx", ["biome", "format", "--write", filePath], {
+    stdio: "ignore",
+    shell: true,
+    timeout: 25_000,
```
james-baker: I wonder how much time this adds to every Edit/Write? Biome is really fast but node might be ~1s?

(and should we lower the timeout tolerance here? I might want to know if it ever takes... 3s?)

.claude/hooks/format-file.mjs
```diff
@@ -0,0 +1,35 @@
...
+  });
+} catch {
+  // Swallow: hook must never block Claude.
```
james-baker: we might want to know, possibly?

.git-blame-ignore-revs
```diff
@@ -0,0 +1,7 @@
+# Revisions listed here are skipped by `git blame` and by GitHub's blame view.
+# Add commits that are pure formatting / mechanical reformatting, never logic
+# changes. To honor this locally:
```
james-baker, line 3: I didn't even know a "blame ignore" feature existed, cool. Do we want the local config line to also go in the dev setup?

lefthook.yml
```diff
@@ -0,0 +1,21 @@
...
+pre-commit:
+  parallel: true
+  jobs:
+    - name: biome
+      glob: "*.{js,jsx,ts,tsx,json,jsonc,css}"
+      # --no-errors-on-unmatched: lefthook may pass paths Biome's `includes`
+      #   excludes; don't fail the commit on those.
+      # --files-ignore-unknown=true: skip files Biome doesn't recognize
+      #   rather than erroring.
+      run: npx biome check --write --no-errors-on-unmatched --files-ignore-unknown=true {staged_files}
```
james-baker: I don't know much about this but, but why does the biome call not itself determine which files it's going to run on? (It looks like it's because lefthook is trying to scope down the biome run to only staged files)

Maybe the "biome doesn't know what's staged, lefthook doesn't know what biome should ignore" clash is just not worth fixing, but it just smells a bit.

scripts/setup.mjs
```diff
@@ -28,9 +42,12 @@ function ensureLocalEnv() {
 function main() {
   console.log("Setting up is-app for local development...\n");
 
-  console.log("1. (and only step currently) Environment file");
+  console.log("1. Environment file");
   ensureLocalEnv();
 
+  console.log("\n2. Git hooks (lefthook)");
```
james-baker, line 68: my dumb question: if it's in package.json, why does it need manual installation?

james-baker (commented):
do more stuff, get asked more questions. :)

nothing blocking

.claude/hooks/format-file.mjs
NorsemanSpiff: We discussed these earlier today, posting comments Claude helped me draft for posterity.

Tested. Numbers from Win11, warm caches, 5 runs each:

| Step | min | p50 | max |
|---|---|---|---|
| `npx biome format <file>` (current) | 1060ms | 1090ms | 2610ms |
| `./node_modules/.bin/biome ...` | ~25ms | ~25ms | ~30ms |
| `node -e '0'` (just Node startup) | 166ms | 166ms | 200ms |

Biome itself is ~25ms. The ~1s/edit cost is npx's bin-resolution overhead — exactly your hunch.

After thinking about it, **dropped the per-edit hook entirely** (98d2fd5). ~1.1s on every Edit/Write is too much for what the other 3 layers (npm test, pre-commit, CI) already catch. Instead added pre-commit telemetry (`scripts/biome-precommit.mjs` → `.scratch/biome-precommit-stats.log`) recording how often a commit actually needed reformatting. If the data says "often," we'll reintroduce the hook with a direct-binary invocation (~200ms, ~5× faster) instead of npx.

So this whole script is gone in 98d2fd5. The follow-up question — "would a fast Claude hook have been worth it?" — gets answered empirically over the next few weeks of pre-commit data.

.claude/hooks/format-file.mjs
NorsemanSpiff: We discussed these earlier today, posting comments Claude helped me draft for posterity.

Moot — the whole script is gone in 98d2fd5 along with the per-edit hook itself (see reply on the timing comment).

For the record, had we kept the hook, the visibility design was going to be exit-non-zero + log file + stdout JSON `additionalContext`, hitting all three audiences (user banner, Claude reactive, debug log). If we ever bring the hook back, that's the pattern.

.git-blame-ignore-revs
NorsemanSpiff: We discussed these earlier today, posting comments Claude helped me draft for posterity.

Done in 98d2fd5 — `scripts/setup.mjs` now runs `git config blame.ignoreRevsFile .git-blame-ignore-revs` on `npm run setup`. Idempotent.

lefthook.yml
NorsemanSpiff: We discussed these earlier today, posting comments Claude helped me draft for posterity.

Real smell, and it's an explicitly documented design trade-off. Biome's [git-hooks recipe](https://biomejs.dev/recipes/git-hooks/) says:

> Note that you don't need to use both `glob` and `--files-ignore-unknown=true`. Using only `--files-ignore-unknown=true` allows handling files supported in the present and in the future by Biome. If you wish more control over which files are handled, you should use `glob`.

We kept both for now. Three filters in play between staged files and what biome touches: lefthook `glob:`, biome's `--files-ignore-unknown=true`, and biome.json `includes` (where we excluded `drizzle/meta` etc.). The glob's main value is "skip the job if zero matches" — avoids the ~1s npx spinup on docs-only commits. Forward-compat downside: if biome adds support for a new file type, the glob silently excludes it until we update.

That gap immediately bit us in dabab1a — the `.mjs` files we shipped in 98d2fd5 (the wrapper script, the modified setup.mjs) weren't matching the glob, so they bypassed pre-commit. Widened the glob to include `mjs,cjs` in that commit. Real instance of the smell you flagged.

Open to flipping to "drop the glob, rely on biome's filtering" if you'd prefer the cleaner single-source-of-truth posture — happy to make that change as a follow-up.

scripts/setup.mjs
NorsemanSpiff: We missed this one earlier today, but it's not a big issue. Posting comment Claude helped me write for posterity.

Not a dumb question at all — I had it wrong yesterday. Lefthook's own `node_modules/lefthook/postinstall.js` already runs `lefthook install` during `npm install` (skipped only when `CI=true`):

```js
if (isEnabled(process.env.CI) && !isEnabled(process.env.LEFTHOOK)) return;
spawnSync(getExePath(), ["install", "-f"], { ... });
```

So our explicit `ensureLefthookInstalled()` is **redundant in the normal dev case**. Kept it anyway for defense-in-depth (pnpm requires `onlyBuiltDependencies` for postinstalls; if someone has `CI=true` locally) and as documentation of intent (a new dev reading `setup.mjs` immediately knows the project uses hooks).

This raises a broader pattern question — *which* devDeps deserve an explicit step in `setup.mjs`? Same question came up for Playwright in #285 (Playwright deliberately doesn't auto-install browsers). Deferred the policy decision to #285 rather than precedent-set here. Added a comment in `setup.mjs` linking #285 + flagging the redundancy. Cross-linking on #285 separately.

benjifriedman:
@NorsemanSpiff we're both working late! I just resolved the homepage merge conflict, everything else looks good

benjifriedman (approved):

NorsemanSpiff:
@benjifriedman thank you good sir! Just merged. Good luck with your late night productivity endeavors!

PR #319, opened 2026-05-30
drop live_desire column (contract step after #245)

oolu4236:
## Summary

Contract step following #245 (quarterly intentions). All `live_desire` data was already migrated to `current_intention` in migration 0011. This PR drops the column.

- `ALTER TABLE profiles DROP COLUMN live_desire` (migration 0012)
- Removed from schema, type definitions, `EDITABLE_PROFILE_FIELDS`, all SELECT statements, test-reset, and test assertions

## Test plan

- [ ] CI lint + functional tests pass
- [ ] `prod:db:expand` run approved after merge

Closes part of #185


james-baker:
Just to correct the test plan: "prod:db:expand run approved after merge" is not true for a contraction, we don't do it, and the way we can tell we don't need it is that the e2e tests pass without running it - they show that the code doesn't rely on whatever we're doing at the db level.

PR #333, opened 2026-06-04
Allow relation value 5

james-baker:
Widens the relationship
  vocabulary to include a fifth tier (5) by extending the accepted range in isRelationValue. Part of expanding the 1-N
  relation scale used by the rating dialog and invite form.


james-baker:
Investigation complete — root cause found (workflow prompt lacks --comment, so the reviewer discards findings). Closing throwaway.

PR #335, opened 2026-06-04
ci: TEMP enable show_full_output on claude-review (diagnostic)

james-baker:
Temporary diagnostic. Surfaces the claude-review transcript so we can see why it posts nothing on PRs. Reverted immediately after observing one PR review. Required check: Lint & Functional Tests (workflow-only change, unaffected).


james-baker:
bypassing for temporary test purposes

PR #371, opened 2026-06-08
test(e2e): make warm-up fail loudly with bounded retries

### LINKED ISSUE #368: e2e cold-start flake recurs: reset.setup.ts warm-up silently swallows its own failures
### Symptom

On PR #367's first e2e run (`deployment_status`-triggered Playwright against the Vercel preview), 1 hard fail + 4 flaky out of 29, every one a `page.goto(…, waitUntil: "load")` exceeding the 30s test timeout, then `net::ERR_ABORTED; maybe frame was detached?` (the abort is Playwright tearing the frame down on timeout). All on `/signin` and `/signup`. A warm rerun passed clean (24→29, 2m39s).

### It's the document, not app code

Pulled the failed run's Playwright traces. The aborted-navigation traces contain **exactly one network request — the document (`/signin`, `/signup`) — and nothing after it**: the serverless function never responded within 30s, so the HTML never parsed and no sub-resources were requested. The same CI run also hit a **100s `actions/checkout` download timeout + backoff**, i.e. a runner-network brownout during that window. Cold function + degraded network → document navigations exceeded 30s.

### Root cause of the *recurrence* (the actual fixable bug)

#360 added a warm-up in `tests/e2e/reset.setup.ts` ("warm up the page-serving path") that hits `/signin`, `/`, `/signup`, `/forgot-password` before timed tests. But it does so via:

```ts
await page.goto(path).catch(() => {});
```

Under the same brownout, those warm-up gotos **also time out and are silently swallowed** — so the routes aren't actually warm when the first real test hits them. #360's protection evaporates in exactly the conditions it exists for, and the failure surfaces downstream as a flaky/failed first test instead of a visible warm-up problem.

### Proposed fix

- Make the warm-up **observably robust**: a couple of retries per route with a longer navigation timeout and `waitUntil: "domcontentloaded"` (lighter than `load`), and **log/annotate on miss** instead of swallowing.
- Consider failing the `setup` project **loudly** if warm-up can't complete after retries — a cold/brownout run becomes a clear setup failure rather than a confusing first-test flake.
- Optionally raise the navigation timeout (or add a navigation retry) for the cold first hits.

### References

- #360 (prior cold-start fix this regresses against), PR #367 (where it recurred)
- `tests/e2e/reset.setup.ts`, `playwright.config.ts` (`retries: 2` in CI, 30s default test timeout)

---
🤖 Filed by Claude Code

james-baker:
## Summary

Replace the e2e warm-up's swallowed `page.goto` calls with a retrying, deadline-bounded warm-up that fails the setup project loudly when a route or the sign-in can't warm.

## Why

#360 added a warm-up in `reset.setup.ts` to pay cold-start cost in untimed setup, but it navigated via `page.goto(path).catch(() => {})`. Under the cold-function + degraded-network brownout it exists to absorb, those gotos time out and are silently swallowed — so the routes aren't actually warm and the failure resurfaces as a confusing first-test flake (recurred on #367).

## Behavior

- Each route warms with one short retry and `waitUntil: "domcontentloaded"` (lighter than `"load"`); the sign-in warm-up retries on the same terms.
- A route or sign-in that can't warm after its retries throws a named error (`warm-up: /signin did not respond ...`), failing the setup project loudly instead of flaking downstream. CI's project-level `retries: 2` still rides out a transient brownout.
- A 75s deadline shared across all routes bounds the worst-case wait: a single hung route gives up at ~38s (was 90s), a full brownout at ~75s (was ~180s), with a 120s test-timeout backstop (was 240s).

## Test Plan

- ran `npm test` locally — lint, typecheck, Vitest functional, and 29 Playwright e2e all green
- confirmed the warm-up setup project warms fast on a healthy run; the new retry/deadline paths only engage on a cold/brownout miss

## Checklist

- [x] N/A — this PR changes no `.claude/skills/**/SKILL.md`.

Closes #368

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/tests/e2e/reset.setup.ts
+++ b/tests/e2e/reset.setup.ts
@@ -29,18 +29,20 @@ const WARMUP_FIRST_NAV_TIMEOUT_MS = 30_000;
 const WARMUP_RETRY_NAV_TIMEOUT_MS = 8_000;
 const WARMUP_ATTEMPTS = 2;
 
-// Hard ceiling on the whole route-warming phase, shared across all routes:
-// once it's spent, warm-up gives up loudly instead of grinding through every
-// remaining route's retries. This — not the per-attempt windows — is what
-// bounds the worst case when the network browns out mid-run. A persistent
-// brownout still recovers via CI's project-level `retries: 2`, which re-runs
-// the whole setup test later.
+// Hard ceilings on each warm-up phase — the routes loop and the sign-in
+// respectively. Once a phase's budget is spent it gives up loudly instead of
+// grinding through every remaining retry. These — not the per-attempt windows
+// — are what bound the worst case when the network browns out mid-run. A
+// persistent brownout still recovers via CI's project-level `retries: 2`,
+// which re-runs the whole setup test later.
 const WARMUP_ROUTES_BUDGET_MS = 75_000;
+const WARMUP_SIGN_IN_BUDGET_MS = 25_000;
 
-// Generous backstop above the route budget + sign-in warm-up, so the named
-// budget/attempt errors below fire first. Playwright's 30s default test
-// timeout is far too tight for a cold first hit, and would trip with an
-// opaque message before the retry logic could run.
+// The setup test timeout sits above both phase budgets plus waitForURL slack
+// (75s + 25s + ~12s), so the named budget/attempt errors below always fire
+// before Playwright's opaque "Test timeout exceeded" message. Its 30s default
+// is far too tight for a cold first hit, and would trip before the retry logic
+// could run.
 const WARMUP_TEST_TIMEOUT_MS = 120_000;
 
 // Navigate to a route with one short retry, bounded by both its per-attempt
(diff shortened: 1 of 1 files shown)


james-baker:
Addressed in 9c016d1 — thanks, this was a real gap, not a false positive.

You're right that the sign-in phase was the one unbounded path. `signInAs`'s `page.goto` and form actions carry no explicit timeout, and Playwright Test defaults `navigationTimeout`/`actionTimeout` to `0`, so they fell back to the remaining test budget — meaning a hung sign-in goto could trip the opaque `Test timeout of 120000ms exceeded` instead of the named warm-up error, exactly as you described. (#368's own symptom — real specs failing as "exceeding the 30s **test** timeout" on unbounded gotos — corroborates the mechanism.)

**Fix:** the sign-in retry loop now runs under its own `WARMUP_SIGN_IN_BUDGET_MS` (25s) deadline. Before each attempt it sets `page.setDefaultNavigationTimeout(remaining)` + `page.setDefaultTimeout(remaining)`, so every operation inside `signInAs` is bounded by the remaining budget (`waitForURL` keeps its explicit 12s). I went with `setDefault` rather than threading a timeout param so the shared `signInAs` helper — which every spec uses — stays untouched.

Worst case is now routes 75s + sign-in 25s + ~12s `waitForURL` slack ≈ **112s**, comfortably under the 120s setup-test timeout, so the named `warm-up: sign-in did not complete …` error always fires first. The `WARMUP_TEST_TIMEOUT_MS` comment is updated to state that relationship.

Verified locally: `npm test` green (lint, typecheck, 385 functional, 29 e2e). The healthy path is unaffected — the budgets never engage when warm-up succeeds normally.

🤖 Posted by Claude Code

PR #395, opened 2026-06-12
feat(skills): vendor upstream skill-creator for team-local access

NorsemanSpiff:
**Summary**: Vendor Anthropic's skill-creator skill at .claude/skills/skill-creator/ so every teammate has it on clone with zero per-user setup. (Source: https://github.com/anthropics/skills/tree/main/skills/skill-creator)

**Why**: We built /commit, /pr, and /ship with skill-creator, but the skill itself wasn't in the repo — each person had to manually mount a per-user copy. That doesn't scale to the team. Vendoring puts it at the path Claude Code auto-loads (`.claude/skills/<name>/`), riding the existing `!.claude/skills/` gitignore negation, with deliberate pinned-SHA updates rather than silent drift.

**Behavior**: `.claude/skills/skill-creator/` is now auto-loaded in every Claude Code session started at the repo root. `scripts/update-skill-creator.mjs` syncs the vendored copy to a new upstream commit and rewrites UPSTREAM.md; `--check` mode compares the pin to upstream and exits non-zero if behind (weekly drift detection in a later PR). The skill remains model-invokable by design — it is a non-side-effecting authoring tool, consistent with the NL-invocation two-tier model.

**Test Plan:**
- ran `npm test` locally — all suites green (typecheck, lint, functional; Playwright not triggered, no app code changed)
- sc-1 (load-and-invoke): `/skill-creator` autocompleted and started intake in a fresh session with no missing-file errors — Blake attested
- sc-2 (validator self-validates): `python .claude/skills/skill-creator/scripts/quick_validate.py .claude/skills/skill-creator` → "Skill is valid!", exit 0 — confirmed locally
- sc-3 (end-to-end authoring): well-formed SKILL.md generated for a throwaway skill; deleted after — Blake attested

**Note:** `.claude/skills/skill-creator/` is excluded from Biome lint (`biome.json` `files.includes` negation) — vendored HTML fails upstream lint rules; this is expected and intentional. The `!.claude/skills/` negation in `.gitignore` that allows team skills to be committed applies here too.

See `docs/plan-skill-creator-vendoring.md` for the full 5-PR strategy (this is PR 1), alternatives rejected, and sequencing notes for the follow-on PRs. Follow-on work tracked in #396.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


james-baker:
> .claude/skills/skill-creator/ is now auto-loaded in every Claude Code session started at the repo root

Looking at the diffs I now question: how many tokens does this skill add to our sessions? Like, <2k, or more like 10k? I hadn't been thinking of this "token budget" cost before.

james-baker:
Oh maybe I asked a bad question. Claude says:

```
Your repo's skills are missing from that list for a good reason: all three (/commit, /pr, /ship) are declared with
  disable-model-invocation: true. That setting means the skill is excluded from the model-visible skill list entirely — and
  that list is exactly what the /context Skills section is measuring (the ~40–360 token per-skill descriptions injected every
  session so the model knows when to fire them). So their standing cost in every session is zero tokens.

  They only cost context at the moment you type the slash command, when the full SKILL.md body gets injected into the
  conversation (it then shows up under "Messages").

Here's the full picture:

  ┌─────────┬───────────────────────────────┬──────────────────────────────┐
  │  Skill  │ Standing cost (every session) │ Cost when invoked (~chars/4) │
  ├─────────┼───────────────────────────────┼──────────────────────────────┤
  │ /commit │ 0                             │ ~3.1k tokens                 │
  ├─────────┼───────────────────────────────┼──────────────────────────────┤
  │ /pr     │ 0                             │ ~2.9k tokens                 │
  ├─────────┼───────────────────────────────┼──────────────────────────────┤
  │ /ship   │ 0                             │ ~2.8k tokens                 │
  └─────────┴───────────────────────────────┴──────────────────────────────┘

  Two notes on the invoked cost:

  - Since /ship orchestrates /pr which orchestrates /commit, a full /ship run ends up loading all three bodies — roughly 9k
  tokens — plus whatever referenced files the skill tells me to read.
  - You can verify empirically: run /context, invoke a skill, run /context again and watch the "Messages" line jump by about
  the skill's size.

  This is the trade disable-model-invocation: true buys you: the skills are free until used, at the price that I can't
  proactively suggest or auto-fire them — which matches their design as explicit check-in workflows. Contrast with the Vercel
  plugin in your /context output: ~32 always-loaded skill descriptions costing ~2.5k tokens every session whether or not you
  ever touch them.
```

If we only pay a context cost when we use skills, then my Q is moot. And I didn't know that "disable-model-invocation" had any impact on context, but it appears to be "just the (small) description, not the skill" so I'm not worried about that.

james-baker (approved):
Nothing in here looks scary or damaging - as part 1 of a multi-part plan, LGTM. I'm curious to learn what all it actually does and how we actually use it to improve and test our skills.

NorsemanSpiff:
Right idea — one nuance: that breakdown covers /commit, /pr, /ship which are all `disable-model-invocation: true`. skill-creator itself is model-invokable, so its description does load every session. Still tiny — ~100 tokens. But it means you can just say "build me a skill that does X" and it fires naturally — which is the point for a tool you use occasionally rather than constantly.

---

NorsemanSpiff:
Near-term: I'm going to run it on /commit, /pr, and /ship to tighten them up — there are rough edges worth fixing and the eval framework makes it possible to actually measure before/after. That's PR 2 in the sequence tracked in #396.

---

PR #397, opened 2026-06-12
feat: add sun/moon theme toggle to top-left header

benjifriedman:
## Summary

Add a sun/moon icon toggle in the top-left corner for quick light/dark switching.

## Why

The existing theme selector lives inside the Settings tab on /me. A top-level toggle provides quicker access without navigating away from the current page.

## Behavior

- A ghost-style icon button appears fixed in the top-left of the viewport (mirrors the hamburger menu's top-right placement).
- Shows a moon icon in light mode, sun icon in dark mode.
- Clicking toggles between explicit light/dark (if the stored pref was "system", the first click resolves to the opposite of the current OS theme).
- Only renders for authenticated users (same as the nav menu).
- Reuses `applyThemePreference` / `readThemePreference` from `src/lib/theme.ts`, so it stays in sync with the Settings tab selector.
- Renders a disabled placeholder pre-hydration to avoid layout shift and hydration mismatch.

## Test Plan

- Biome check passes on both changed files
- TypeScript diagnostics clean on both files
- Manual: toggle switches between light/dark, persists across reload, stays in sync with /me#settings selector

## Checklist

- [ ] If this PR changes `.claude/skills/**/SKILL.md`, smoke-test the affected Skill on a realistic prompt.

### DIFF
--- a/src/components/site-header.tsx
+++ b/src/components/site-header.tsx
@@ -1,13 +1,15 @@
 "use client";
 
-import { House, Menu, ShieldCheck } from "lucide-react";
+import { House, Menu, Moon, Settings, ShieldCheck, Sun, XIcon } from "lucide-react";
 import Link from "next/link";
 import { usePathname } from "next/navigation";
+import { useEffect, useState } from "react";
 
 import { useAuth } from "@/components/auth-provider";
 import { Button, buttonVariants } from "@/components/ui/button";
 import { Sheet, SheetClose, SheetContent, SheetHeader, SheetTitle, SheetTrigger } from "@/components/ui/sheet";
 import { clearNavHistory } from "@/lib/route-labels";
+import { applyThemePreference, readThemePreference, type ThemePreference } from "@/lib/theme";
 import { cn } from "@/lib/utils";
 
 // Internal nav links share one shape: a SheetClose (so the menu closes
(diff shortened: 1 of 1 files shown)


james-baker:
With my designer hat on, I'll vote no on the placement - I think we might put a "Home" icon there, but I don't think this function deserves that real estate.

What I'd be interested in is, can this same sun/moon icon toggle UI be put inside the slide-out menu? On the "Menu" or Name lines, maybe? I'm not exactly sure on the find details, but that feels like "the place" generally to me.

james-baker:
Second question: How do people deal with the 3-state-ness of light v dark v *system*?

benjifriedman:
@james-baker Totally, fair enough on placement / real estate. I'll make an update to move the toggle to the sidebar for you to check out. 

On the 3-state question: the existing theme infra you built already handles this. The default is "system" (auto-follows OS, including live-tracking if the OS pref changes mid-session). The toggle just lets someone quickly flip to explicit light or dark. Once they've clicked it they're on a manual override, but they can always reset to "system" from /me#settings.

So this PR isn't adding new theme logic — just surfacing a quicker toggle for the light/dark switch that reuses applyThemePreference / readThemePreference from  theme.ts

benjifriedman:
Here's an idea that's in commit [b297cb3](https://github.com/Intentional-Society/is-app/pull/397/commits/b297cb398ffcf2edc82157798b3640aec9513a58), probably an improvement, but still doesn't feel totally right. I would just worry that it's starting to clutter the sidebar a bit. Maybe could go above Home, or what do you think? I might be overthinking this 

Screenshot 2026-06-12 at 11 58 04 AM
Screenshot 2026-06-12 at 11 57 58 AM

james-baker:
Yeah I agree it doesn't go there. It's not a page. 

Maybe:
`
Menu <gear> <sun>     X
`
on that top line? We could also link the user's name to the profile section of their "My page" and cut that link from the menu.

Oh, and I notice I'm on team "show the current state" re this icon, when it's just an icon, as opposed to "show the opposite state it could be". (In light mode, my brain wants to see sunshine, in dark mode my brain wants to see the moon)  (While if the 'button/label text' is included, then it should show the opposite) WDYT?

james-baker:
By the way #401 just added a Home icon.

benjifriedman:
Agreed. I think this is much better:
Screenshot 2026-06-12 at 9 59 05 PM
Screenshot 2026-06-12 at 9 59 18 PM

james-baker:
I like it!

Issue #350, opened 2026-06-05
Admin hint form gives no feedback when a create is a no-op (created: false)


james-baker:
**What**

In the admin Web section, creating a hint (relator → target) silently "succeeds" with no visible result when the pair already has a relations row. The form clears as if it worked, but no hint was created. An admin can retry the same pair repeatedly with no signal (observed on prod).

**Why it happens**

`createRelationHint` uses `onConflictDoNothing` on the `(relatorId, relateeId)` primary key and returns `{ ok: true, created: false }` when a row already exists (src/server/relations.ts:491-503). The admin UI treats any `res.ok` as success — `onSuccess` clears the form and refetches — so `created: false` looks identical to `created: true` (src/app/admin/admin-hints.tsx:37-55).

When the existing row is a real (rated) relation (`isHint = false`), the no-op'd pair also never appears in the Pending hints list, which filters `isHint = true` (src/server/relations.ts:534). So the admin gets zero feedback.

**Proposed fix**

Surface the `created` flag in the UI. On `created: false`, show an info notice instead of silently clearing — e.g. "Already related or already hinted — nothing to add." Client-only change in admin-hints.tsx; the response already carries `created`.

**Acceptance**

- Create a hint for a pair with no existing relation → success as today.
- Create a hint for a pair that already has a relation (hint or rated) → visible no-op notice so the admin understands nothing was created.

🤖 Filed by Claude Code

oolu4236:
Already resolved on `main` by #405 (commit 4624d02): `admin-hints.tsx` now surfaces the `created` flag — on `created: false` it shows "Already related or already hinted — nothing was added." instead of silently clearing. Closing as done.

Issue #346, opened 2026-06-04
Member typeahead near the page bottom scrolls the viewport on open, hiding the trigger


james-baker:
## Symptom

On `/invites` (and anywhere the member typeahead is used), opening the "Add a connection recommendation…" control while it sits near the bottom of the viewport scrolls the page on open: the autocomplete opens *above* the trigger, then the viewport jumps so its bottom edge aligns with the bottom of the popover — pushing the trigger you just clicked out of view.

Shared component: `src/components/member-typeahead.tsx` (base-ui `Popover` + cmdk `Command`). Also affects the admin-hints typeahead, which reuses it.

## Root cause (two independent scroll sources)

1. **base-ui focus.** `FloatingFocusManager` focuses the first tabbable child (the cmdk search input) on open, and only passes `preventScroll: true` when it focuses the popup *container*, not a child:
   ```
   enqueueFocus(elToFocus, { preventScroll: elToFocus === floatingFocusElement, ... })
   ```
   Focusing the input therefore scrolls the page to reveal it.
2. **cmdk.** On mount cmdk selects the first item and calls `item.scrollIntoView({ block: "nearest" })`, which also scrolls ancestor scrollports (the page) when the popover doesn't fully fit the viewport.

## Attempts that did NOT fix it

- Overriding base-ui `initialFocus` to focus the input ourselves with `{ preventScroll: true }` — addressed source (1) only; cmdk still scrolled.
- Adding 300px of blank space below the page content — no effect.

## Proposed fix

Cap the popover popup to base-ui's `--available-height` (exposed by `useAnchorPositioning`) so it always fits the viewport — the idiom base-ui's own Select/Combobox use. When the popup fits, neither the focus call nor cmdk's `scrollIntoView` needs to move the page. Likely `max-height: var(--available-height)` on the popup with the cmdk list flexing/scrolling inside.

## Repro

1. Sign in, go to `/invites`.
2. Make the window short enough that the form scrolls; scroll so the "Add a connection recommendation…" trigger is near the viewport bottom.
3. Click it → the page scrolls and hides the trigger.

🤖 Filed by Claude Code

oolu4236:
Already resolved on `main` by #405 (commit 4624d02): the member-typeahead popover is now bounded to the viewport (`maxHeight: var(--available-height)` + a scrollable `CommandList` capped at `min(300px, available-height-52px)`), so opening it near the page bottom no longer scrolls the trigger out of view. Closing as done.

PR #405, opened 2026-06-14
fix: admin QueryClient 500, no-op hint feedback, typeahead viewport scroll

### LINKED ISSUE #400: Local dev: /admin 500s — AdminHints SSR crashes with 'No QueryClient set'
Local dev (`npm run dev`, Turbopack) serves a 500 for `/admin`, failing `admin.spec.ts` ("admin can open the /admin hub without console errors") on every local e2e run. SSR of the page crashes in `AdminHints`:

```
Error: No QueryClient set, use QueryClientProvider to set one
    at AdminHints (src\app\admin\admin-hints.tsx:30:37)
  digest: '1292750205'
```

Evidence so far:

- Deterministic, not a flake — reproduced across three runs (full suite + two isolated `npx playwright test tests/e2e/admin.spec.ts`), on a freshly started dev server.
- Reproduces on unmodified `main` (verified via `git stash` A/B on 2026-06-12, main at 903b2e8), so no open branch is the cause.
- `/admin/programs` renders 200 in the same runs; only `/admin` (the page composing `AdminHints`) fails.
- Provider nesting is correct: root layout wraps `children` in `QueryProvider`, and the same code passes e2e in CI against Vercel preview (production build). That makes this look like a dev-build/Turbopack module-duplication issue (two copies of `@tanstack/react-query` → context identity mismatch) rather than an app bug — unverified.

Until fixed, local `npm test` cannot go fully green; CI (Vercel preview) is unaffected.

🤖 Filed by Claude Code

oolu4236:
## Summary

Three small bug fixes:

- **#400** — `/admin` 500s in local dev (Turbopack). `AdminHints`, `AdminHidden`, and `ButtondownSyncButtons` are client components that call `useQuery`, but Turbopack's module deduplication can create a context identity mismatch. Wrapping them in a local `QueryProvider` on the admin page fixes it without affecting CI (production build was unaffected).

- **#350** — Admin hint form silently reset when the pair already existed (`created: false`). Now shows "Already related or already hinted — nothing was added." The form stays populated so the admin knows what happened.

- **#346** — Member typeahead near the bottom of the viewport scrolled the page on open. Fixed by capping `PopoverContent` to `--available-height` and constraining `CommandList` scroll inside, so neither base-ui's focus call nor cmdk's `scrollIntoView` needs to move the page.

## Test plan

- [ ] Sign in as admin, go to `/admin` — page loads without 500
- [ ] Try creating a hint for Amara Diallo → Beatriz Costa (already hinted) — see no-op notice
- [ ] Go to `/invites`, make the window short, scroll so typeahead is near bottom, open it — no viewport jump

Closes #400, #350, #346


benjifriedman (approved):

james-baker (approved):
Looks fine to me, though I'll note that the nested QueryProvider thing is still just a guess. (I assume you didn't manage to replicate it and observe it change in response to the fix-attempt, is that true?)

oolu4236:
> Looks fine to me, though I'll note that the nested QueryProvider thing is still just a guess. (I assume you didn't manage to replicate it and observe it change in response to the fix-attempt, is that true?)

True, I wasn't able to reliably reproduce the 500 locally to confirm the fix. The QueryProvider wrapping follows the same pattern used on other pages that don't hit the issue. if it comes back I'll dig further.

james-baker:
> True, I wasn't able to reliably reproduce the 500 locally to confirm the fix. The QueryProvider wrapping follows the same pattern used on other pages that don't hit the issue. if it comes back I'll dig further.

Yeah that's fine, not an issue

PR #407, opened 2026-06-14
refactor: convert Sentry namespace imports to named imports (#256)

### LINKED ISSUE #256: Enable Biome performance/noNamespaceImport and convert the seven existing namespace imports
Standardize how third-party modules like `@sentry/nextjs` are imported. Seven files in the repo currently use `import * as Sentry from "@sentry/nextjs"`; named imports would tree-shake universally, but the namespace form keeps a readable `Sentry.X` prefix at call sites. Pick one of the two options below and apply it consistently.

**Current state — seven namespace imports, all `@sentry/nextjs`:**

| File | Names actually used |
|---|---|
| `instrumentation.ts` | `captureRequestError` |
| `instrumentation-client.ts` | _check at conversion time_ |
| `sentry.server.config.ts` | `init` |
| `src/app/global-error.tsx` | `captureException` |
| `src/server/programs.ts` | `captureException` |
| `src/app/auth/callback/route.ts` | `captureException` |
| `tests/functional/server/programs.test.ts` | `captureException` (mocked via `vi.mocked`) |

### Option A — aliased named imports, enable the Biome rule

Convert every namespace import to named imports with a third-party-name alias, so call sites still read with a prefix:

\`\`\`ts
import {
  captureException as sentryCapture,
  withScope as sentryWithScope,
} from "@sentry/nextjs";

sentryCapture(err);
\`\`\`

Enable the Biome rule so the codebase can't drift:

\`\`\`json
"linter": {
  "rules": {
    "performance": {
      "noNamespaceImport": "error"
    }
  }
}
\`\`\`

**Pros**
- Tree-shaking guaranteed across every bundler, present and future.
- Lint rule enforces the standard — new code can't regress.
- Keeps a 3rd-party-name prefix at call sites for readability (`sentryCapture` reads as Sentry-y).

**Cons**
- Verbose at the import block; aliases compound if a file uses many names from the same SDK.
- Cosmetic mismatch with the official Sentry docs, which use namespace imports throughout.

### Option D — keep namespace imports, write a code-style note

Leave the seven files as-is. Don't enable the Biome rule. Document the convention in `docs/` (probably a short addition to an existing doc rather than a new file): **named imports by default; namespace imports allowed for SDK-style packages where the prefix aids readability (`Sentry.`, `z.`, etc.).**

**Pros**
- `Sentry.captureException(err)` reads cleanly at call sites — the package source is obvious without aliasing.
- Modern bundlers (Webpack 5, Rollup/Vite, esbuild, Turbopack) track property access on namespace imports of ESM packages and tree-shake the unused surface, so the practical bundle-size cost is near zero for `@sentry/nextjs`.
- Zero churn in the codebase today.

**Cons**
- Convention lives in docs, not in the linter — easy to drift if a contributor hasn't read the doc.
- Tree-shaking effectiveness depends on bundler + the package being ESM; future SDK / bundler combos might not preserve the property.
- Allowing exceptions opens \"is package X SDK-style enough?\" debates per import.

## Apply consistently

Whichever option wins, apply to all seven files (and going forward, to any new third-party module that lands in the repo) — half-and-half is the worst outcome.

---

🤖 Filed by Claude Code

oolu4236:
## Summary

Replaces all 11 `import * as Sentry from "@sentry/nextjs"` with named imports. Only 5 distinct names are used across the codebase: `captureException`, `captureMessage`, `captureRequestError`, `captureRouterTransitionStart`, `init`, `replayIntegration`. Named imports are recognizable without a namespace prefix.

Enables `performance/noNamespaceImport` in `biome.json` to prevent drift.

## Test plan
- [ ] CI lint passes (Biome rule now enforced)
- [ ] No runtime regressions (Sentry still captures errors)

Closes #256

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -45,6 +45,9 @@
     "enabled": true,
     "rules": {
       "recommended": true,
+      "performance": {
+        "noNamespaceImport": "error"
+      },
       "style": {
         "noNonNullAssertion": "error"
       },
--- a/instrumentation-client.ts
+++ b/instrumentation-client.ts
@@ -1,4 +1,4 @@
-import * as Sentry from "@sentry/nextjs";
+import { captureRouterTransitionStart, init, replayIntegration } from "@sentry/nextjs";
 
 import { scrubClientEvent } from "@/lib/sentry-scrub";
 
@@ -18,7 +18,7 @@ const debugReplay = (() => {
   }
 })();
 
-Sentry.init({
+init({
   dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
   // Report from production deploys only: preview e2e runs were sending
   // thousands of tunnel POSTs per run (Vercel traffic-spike alerts), and
@@ -30,7 +30,7 @@ Sentry.init({
   replaysOnErrorSampleRate: debugReplay ? 1.0 : 0,
   integrations: debugReplay
     ? [
-        Sentry.replayIntegration({
+        replayIntegration({
           maskAllText: true,
           blockAllMedia: true,
         }),
@@ -39,4 +39,4 @@ Sentry.init({
   beforeSend: scrubClientEvent,
 });
 
-export const onRouterTransitionStart = Sentry.captureRouterTransitionStart;
+export const onRouterTransitionStart = captureRouterTransitionStart;
--- a/instrumentation-client.ts
+++ b/instrumentation-client.ts
@@ -1,4 +1,8 @@
-import { captureRouterTransitionStart, init, replayIntegration } from "@sentry/nextjs";
+import {
+  captureRouterTransitionStart as Sentry_captureRouterTransitionStart,
+  init as Sentry_init,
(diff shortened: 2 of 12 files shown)


james-baker (changes requested):
The way this was implemented here is actually neither of the two options listed in #256. I think I'm a fan of A over D (and this is closer to A), so let's go that way: - adding the 3rd party namespace to the named imports with "as" aliasing:

import { captureException as Sentry_captureException }

oolu4236:
> The way this was implemented here is actually neither of the two options listed in #256. I think I'm a fan of A over D (and this is closer to A), so let's go that way: - adding the 3rd party namespace to the named imports with "as" aliasing:
> 
> import { captureException as Sentry_captureException }

Updated to use the Sentry_ alias pattern (e.g. import { captureException as Sentry_captureException }) instead of namespace imports. I also removed unused imports from the test files, since those modules are already mocked with vi.mock.

james-baker (approved):
LGTM

PR #404, opened 2026-06-14
feat: enhanced program descriptions and status indicators (#403)

oolu4236:
## Summary

- **Two-level descriptions**: new `blurb` field (short card tagline, max 200 chars) + existing `description` kept as full detail on the program page. Falls back to `description` on the card if no blurb set. Admin edit form gains separate fields for both.
- **Status indicators**: green pulse (ongoing, has members), amber pulse (gearing up, signups open), grey static (ready). Member avatar facepile (up to 5 faces) on active program cards.

## Test plan

- [ ] Go to `/admin/programs` → open a program → fill in Blurb and Full description → save
- [ ] Go to `/programs` → card shows blurb + correct status dot + member avatars
- [ ] Click through to detail page → full description shown
- [ ] Program with no members + signups open → amber dot
- [ ] Program with no members + signups closed → grey dot

Closes #403

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #408, opened 2026-06-14
feat: remove password option (#212)

### LINKED ISSUE #212: "Remove password" option — let members revert to email-only sign-in
Spun out of the #200 review. #200 adds a set/change-password section to `/profile/edit`; the natural complement is a way to *remove* a password and go back to magic-link-only sign-in.

Filing this separately because — unlike the label tweaks in #200 — it's a backend feature with real feasibility constraints, surfaced below.

## Goal

A member who has set a password can remove it, reverting to email/magic-link sign-in only. The control would be a "Remove password" button in the change-password section, enabled only when the account currently has a password.

## Feasibility — checked, not punted

Two halves, both lacking first-class Supabase support:

**1. Detecting whether the account has a password** — needed to enable/disable the button.
- Supabase exposes no `has_password` field on the user object.
- The `auth.users.encrypted_password` column is *not* a reliable signal — magic-link signups can populate it; it is reportedly empty only for OAuth users. So `encrypted_password IS NULL` does not mean "no password."
- Workable path: track it ourselves. Add a `hasPassword` boolean (best on the `profiles` table, which we already own) and set it `true` wherever a password is set — the welcome form and #200's change-password form. More robust than a `user_metadata` flag.

**2. Removing the password** — the actual action.
- There is no Supabase API — client or admin — to clear a password while keeping the account.
- The only mechanism is an unsupported direct mutation of `auth.users` (we hold a superuser Postgres connection), which reaches into GoTrue internals, is fragile across Supabase upgrades, and is semantically murky — nulling `encrypted_password` is not even the documented magic-link state.

## Recommendation

The detection half is cleanly solvable app-side; the removal half has no supported path. So step one is a decision: is "remove password" worth an unsupported `auth.users` write, or do we wait for a real Supabase API? If we proceed, the scope is:
- `profiles.hasPassword` column + migration; set it on every password-set path.
- A server endpoint (Hono, admin/superuser) to perform the removal.
- The enabled/disabled "Remove password" button in #200's section.

Does not block #200 — that ships the set/change flow as-is.

oolu4236:
## Summary

Adds a self-serve "Remove password" button to `/me` settings, shown only when `profiles.has_password` is true.

**Detection** — `profiles.has_password` boolean (migration 0014, default false). Set to `true` via `POST /me/password-flag` whenever a password is successfully set via `ChangePasswordForm`.

**Removal** — `DELETE /me/password` replaces the password with an unguessable `crypto.randomUUID()` via the Supabase admin API (`auth.admin.updateUserById`), then clears the flag. The member is forced back to magic-link sign-in. This uses the supported admin API path rather than writing directly to `auth.users`.

**UI** — `RemovePasswordForm` shows idle → confirm → removing states. Only rendered when `profile.hasPassword` is true.

## Test plan
- [ ] Set a password at `/me#settings` → "Remove password" button appears
- [ ] Click "Remove password" → confirm → button disappears
- [ ] Sign out, attempt password sign-in → fails (forced to magic link)
- [ ] Set a new password → "Remove password" reappears
- [ ] `prod:db:expand` approval needed for migration 0014

Closes #212


drizzle/0014_mean_crusher_hogan.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "profiles" ADD COLUMN "has_password" boolean DEFAULT false NOT NULL;
```
james-baker: I don't see anything that backfills this or figures it out for existing users with existing passwords, so won't this be wrong for anyone who already has a password?

src/server/profiles.ts
```diff
@@ -519,6 +522,23 @@ export const reactivateProfile = async (userId: string): Promise<{ ok: true } |
...
+// Removes the member's password by replacing it with an unguessable random
+// value, forcing them back to magic-link / email sign-in. Uses the admin
+// API which is the only supported server-side path; direct auth.users
+// mutation is intentionally avoided.
+export const removePassword = async (userId: string): Promise<{ ok: true } | { error: string }> => {
+  const { error } = await supabaseAdmin.auth.admin.updateUserById(userId, {
+    password: crypto.randomUUID(),
+  });
+  if (error) return { error: error.message };
+  await db.update(profiles).set({ hasPassword: false }).where(eq(profiles.id, userId));
```
james-baker: this should call setPasswordFlag instead of duplicating the implementation

drizzle/meta/_journal.json
```diff
@@ -99,6 +99,13 @@
       "when": 1780506291419,
       "tag": "0013_add_profile_deactivated_at",
       "breakpoints": true
+    },
+    {
+      "idx": 14,
```
james-baker, line 104: 14 gets restamped to 15 now?

james-baker (changes requested):

drizzle/0014_mean_crusher_hogan.sql
oolu4236: Added a backfill to the migration so existing users who already have a password are automatically marked as having one.

src/server/profiles.ts
oolu4236: Updated `removePassword` to reuse `setPasswordFlag` instead of repeating the same database update logic.

drizzle/meta/_journal.json
oolu4236: Renumbered the migration to `0015` because the program descriptions work (#404) now uses `0014`.

james-baker (approved):
changes look good

james-baker:
The "Bug: RemovePasswordForm gets stuck on "Removing password…" after a successful deletion" code review finding looks serious enough to fix IMHO

Issue #396, opened 2026-06-12
Track skill-creator vendoring follow-on work (plan PRs 2–5)


NorsemanSpiff:
### Problem

PR #133 established repo-owned Claude Skills (`/commit`, `/pr`, `/ship`) built and maintained via Anthropic's upstream `skill-creator` workflow — but skill-creator itself wasn't in the repo, so each maintainer had to manually wire it into their session. Not repeatable.

### Status

Investigation and approach are decided — see `docs/plan-skill-creator-vendoring.md` (5-PR sequence, alternatives rejected, decision log). This issue tracks the remaining work.

- **PR 1 — shipped in #395:** skill-creator vendored verbatim at `.claude/skills/skill-creator/`, pinned to an upstream SHA; refresh/check via `scripts/update-skill-creator.mjs`; prerequisites, commands, and artifact policy in `docs/doc-skill-creator.md`; generated output gitignored; acceptance evals (`evals/skill-creator.evals.json`) run and captured in the PR.
- **Prerequisite cleared (2026-06-26):** Plan PR 2 was sequenced behind the NL-invocation revision, which has now **landed** (#353/#433/#459/#460) — so **PRs 2–3 are unblocked.** The plan's verified post-NL state (eval counts, line counts, `/commit` Step 0 + approval-block locations, `settings.json` tracked) is captured in `docs/plan-skill-creator-vendoring.md` via #468.

### Remaining

- [ ] **Plan PR 2 — structural gate:** Vitest test asserting the team-skill contract for `{commit, pr, ship}` (frontmatter, section order, per-skill invocation policy, ≥3 evals each). Note: the repo's own gate is deliberately the contract, *not* upstream's `quick_validate.py` (known divergence: it rejects `disable-model-invocation`). ~~Sequenced after the NL-invocation revision lands.~~ **Unblocked 2026-06-26** — that revision landed (#353/#433/#459/#460); read the verified post-NL state in `docs/plan-skill-creator-vendoring.md` rather than re-deriving.
- [ ] **Plan PR 3 — drift detection:** weekly workflow running `update-skill-creator.mjs --check`, opening/updating a tracking issue when behind; `/commit` surfaces the acceptance evals when the vendored directory is in the payload.
[...]

PR #133, opened 2026-05-08
Spec: portable AI procedure framework v5

NorsemanSpiff:
## Summary

v5 of the portable AI procedure framework spec — ships `docs/spec-portable-ai-procedures.md` as the canonical design RFC addressing all feedback from James Baker's 42-comment review. No code changes; docs-only addition.

@james-baker — this revision is ready for your review. A separate dev-plan PR will follow shortly with the implementation plan for building the three Skills.

## What James asked for

James's review flagged four primary concerns across 42 comments:

- **Scope creep** — the original RFC described adapter architecture for Codex, Copilot, and Aider alongside Claude Code, adding maintenance surface without v1 value.
- **Payload safety** — `/commit` had no deterministic staging, no suspicious-file blocker, and no single-checkpoint approval model.
- **Missing protocols** — no AI co-author trailer format, no schema expand-contract gate, no clear CI-wait policy for `/ship`.
- **Complexity** — too much left to per-invocation prompting; branching logic (dirty state, existing PR, docs-only) needed explicit step-list dispatch.

## What v5 changes

**Skills-only direction (C.1 — resolved)**
- v5 canonicalizes everything in `.claude/skills/{commit,pr,ship}/SKILL.md`. No `docs/ai-procedures/` tree, no index file, no AGENTS.md, no adapter directories, no `.claude/commands/*`.
- `CLAUDE.md`, `README.md`, and four strategy docs each get lightweight back-links; all procedure text lives in the SKILL.md bodies.
- Each Skill sets `disable-model-invocation: true` — explicit slash invocation only; no natural-language triggers.

**`/commit` refuses combined expand+contract payloads (C.2 — James confirmed 2026-05-23)**
- `/commit` Step 9 refuses a single commit that mixes expand and contract schema changes. This is defense-in-depth beyond James's literal ask; James confirmed the interpretation on 2026-05-23.
- Expand-phase commits: schema deploy dispatched via `npm run prod:db:expand` (workflow-dispatcher with maintainer approval in the `prod-db` environment) after PR creation; `/ship` requires the post-deploy e2e run to pass before merging.

**PR #236 merge-method and branch-protection facts**
- Repo allows only merge-commit (`allow_squash_merge: false`, `allow_rebase_merge: false`). `/ship` always uses `gh pr merge --merge --delete-branch`; no `--squash` flag exists.
[...]

PR #433, opened 2026-06-18
feat(skills): natural-language invocation for /commit and /pr (#353)

### LINKED ISSUE #353: Allow 0-to-3 skills to be model-invoked
If you ask Claude to commit or PR or ship, it first says "I'll run the XYZ skill" and then it gets blocked:

"Error: Skill pr cannot be used with Skill tool due to disable-model-invocation"

It's noticing that "disable-model-invocation" (presumably a setting/configuration name somewhere) that has me wondering if we can-and-should loosen some of these up to allow model invocation.

Pro: I fairly often want to say something like "do XYZ and then PR" or "if XYZ, then PR" and it seems like this could enable smoother flow in those cases.

Con: this is a form of gate, is it intentional to try to ban the models from deciding they know enough user intent to execute it on their own decision?  If so, I notice my Claude says "if you'd rather I just do it manually right now (run npm test, show you the commit message, push, open the PR via gh), say the word and I'll take it from there" and I could easily imagine that "disable-model-invocation" doesn't actually stop a confident agent from "working around" the restriction.

Maybe we allow it for commit and PR but not ship, based on an arbitrary amount of caution/intentionality we want?

NorsemanSpiff:
Summary: /commit and /pr become invocable by natural language (not just the slash
command), gated by a one-tap Step-0 intent confirmation with a per-machine opt-out;
/ship stays explicit-only and gains a checked-in `gh pr merge` permission ask-rule as
a harness-level merge guard.

Why: Per #353, when contributors expressed commit/PR/ship intent in natural language,
agents either errored out (disable-model-invocation) or approximated the workflow
outside the official Skills, bypassing the team guardrails. This routes NL intent
through the real Skills, tuned by risk: the reversible actions (/commit, /pr) are
NL-invocable with a light confirmation; the irreversible one (/ship, merge-to-prod)
stays explicit-only and is backstopped by a harness permission prompt. Full design and
decision record: docs/plan-skill-nl-invocation.md.

Behavior:
- /commit and /pr: removed `disable-model-invocation`, rewrote descriptions for NL
  matching, added a model-invoked Step 0 intent gate. Step 0 fires only on NL
  invocation and is skipped on a verified slash entry, a live delegation marker, or the
  opt-out file. Options: Proceed / Proceed-and-don't-ask-again / Stop.
- "Fires once, then never": the don't-ask-again option writes the gitignored
  .claude/skip-nl-confirm-commit-pr.local, so experienced contributors clear the
  confirmation in one keystroke and never see it again; new contributors keep the check.
- Delegation marker (.claude/.nl-delegation-active, `<parent>\t<ISO>`): parent writes it
  immediately before delegating and deletes it after the call returns; the callee
  consumes it at Step 0 (honored only if <30s old) — a lease + release-in-finally so a
  crashed delegation can't leave a stale marker that suppresses a later standalone run.
- /ship: keeps disable-model-invocation (explicit-only). New checked-in
  .claude/settings.json adds an `ask` rule on Bash/PowerShell `gh pr merge *` — a human
  approves any merge in every session, un-weakenable by a local allow or bypass
  (deny->ask->allow, first match). /ship now prints a required PR#/title/checks line
  before the merge command.
- Docs swept off "explicit-only" to the two-tier policy (dated v1.1): CLAUDE.md,
  strategy-committing.md (contributor how-to), strategy-security.md, spec-portable-ai-procedures.md.
- evals/evals.json: +4 NL-routing cases (commit-4, commit-5, pr-8, ship-4).

Test Plan:
- ran `npm run lint` (biome) - clean
- ran `npm run test:functional` - 502/503; the one failure is sync-locks.test.ts (a
  DB-lease test untouched by this PR) which passes 7/7 in isolation = pre-existing
  full-suite flake, not a regression
- live verification in fresh sessions: NL "commit this" -> /commit Step 0 fires first,
  Stop = zero side effects; slash /commit -> Step 0 skipped; opt-out file present ->
  Step 0 skipped; NL "open a PR" -> /pr Step 0 fires first - all pass
- anti-simulation: a fresh session asked to run `gh pr merge` directly refused and
  redirected to /ship (model-level guardrail confirmed)

Follow-ups (not blockers):
- The harness ask-rule's full proof on the actual /ship merge path can only be exercised
  once a real PR exists, so /ship's conversational Y/n is RETAINED here (ask-rule
  additive); removing it is a documented fast-follow after the proof runs against this PR.
  This supersedes the PR #133 step-10 "no second prompt" decision - flagged for review.
- Two pre-existing /commit guardrail evals (contract-phase, suspicious-file false-positive)
  are deferred to #396 (comment filed).

Reviewer notes:
- The .gitignore change re-includes `!.claude/settings.json` past the `.claude/*`
  exclusion - intentional (the checked-in permission rule must ship), same pattern as
  `!.claude/skills/`.
- @james-baker: this changes the /ship step-10 confirmation model you set in #133 -
  please sanity-check the two-tier rationale and the retained-Y/n fast-follow.

Closes #353


james-baker (approved):

james-baker:
All looks good to me! (I didn't read the 500 line spec, but I read everything else)

NorsemanSpiff:
**Review comment follow-up (prior session).** The H1 heading flag at `docs/strategy-security.md:185` is a false positive — the file uses H1 for all top-level sections (`# HTTP Headers`, `# Supply-chain quarantine`, `# Secret rotation`), with H2 for subsections; `# Agent merge guardrail` follows that convention correctly. Also fixed a pre-existing missing trailing newline in the same file.

---

PR #439, opened 2026-06-19
test(e2e): warm authed /welcome routes in setup (#437)

### LINKED ISSUE #437: e2e flake: cold /welcome routes time out completeWelcome — warm them in reset.setup.ts
## Summary

The e2e suite intermittently goes red when the Vercel preview is cold, because the welcome-flow routes (`/welcome`, `/welcome/agreements`, `/welcome/profile`, `/welcome/programs`) are **not** in the warm-up set. The first tests that walk the flow via the `completeWelcome` helper pay the cold-compile cost and blow the 12s `waitForURL` ceiling. This is the same cold-start flake family fixed earlier in #360 / #368 / #371 — the warm-up just doesn't cover this segment yet.

Surfaced on the #433 e2e run ([job](https://github.com/Intentional-Society/is-app/actions/runs/27775034658/job/82184866115)): **1 failed, 4 flaky, 31 passed**. That PR is skills/docs-only and changes no runtime code, so the failures are environmental, not a regression.

## Evidence

Every failure is the same signature — `TimeoutError: page.waitForURL: Timeout 12000ms exceeded` inside `completeWelcome` (`tests/e2e/helpers/session.ts:75/81/96`), never in test-body assertions. The error-context navigation logs show the page *was* moving through `/welcome/*`, just too slowly (with redirect races against the step-gating).

Timing from one trace (`myweb.spec.ts:28`), three navigation waits in the same flow:

| Step | State | Duration |
|------|-------|----------|
| sign-in → `/welcome` | warm | 0.56s |
| `/welcome` → `/welcome/agreements` | warm | 0.28s |
| click "I agree" → `/welcome/profile` | **cold** | **12.00s → TIMEOUT** |

The "I agree" transition (a server action that persists the agreement and redirects to `/welcome/profile`) hit the full ceiling while the preceding navigations in the same flow took under a second. That's a cold server-action / route-segment hit.

## Root cause

`reset.setup.ts` only warms:

```js
const WARMUP_ROUTES = ["/signin", "/", "/signup", "/forgot-password"];
```

plus one authed sign-in (which warms the landing, not the welcome segment). The `/welcome/*` route-group function is therefore cold at suite start; the first `completeWelcome` walk eats the cold compile and times out. The four "flaky" cases recovered once the segment warmed on retry; `myweb.spec.ts:57` exhausted its retries, so the job went red.

## Proposed fix

Warm the welcome segment in `reset.setup.ts` so its function is hot before the suite runs — consistent with the prior cold-start fixes. Two notes for whoever picks this up:

- **Authed + GET-only.** `/welcome/*` requires auth, so a plain (unauthenticated) `warmRoute` goto would just bounce to `/signin`. Warm it on the already-signed-in page from the sign-in warm-up. **Do not submit the steps** — completing the flow sets the welcome markers and the first real test would no longer see a fresh user (the whole point of the reset). A GET of each step page compiles the segment bundle (incl. its server action) without advancing state.
- **Hold the line** on the prior fix's constraints: don't re-inflate `TIMEOUT_MS` and don't swallow the timeout (per #360/#368/#371). The route should be warm, not the ceiling higher.

## Refs

- Prior cold-start fixes: #360, #368, #371
- Related (different mechanisms): #358 (concurrent runs clobber shared DB), #295 (welcome-gate read path probe)

james-baker:
## Summary

Warm the authed `/welcome` step routes in the e2e setup project so the first `completeWelcome` walk no longer pays cold-render cost inside its 12s `waitForURL`.

## Why

`reset.setup.ts` warmed the public routes and the authed landing, but not `/welcome/*`. On a cold Vercel preview the `/welcome/profile` and `/welcome/programs` RSC renders compile lazily, so the agreements → profile hop blew `completeWelcome`'s 12s `waitForURL`. This surfaced on #433's e2e run as 1 failed + 4 flaky — every failure the same `waitForURL` timeout inside `completeWelcome`, none in test-body assertions.

## Behavior

A third warm-up phase GETs `/welcome → agreements → profile → programs` on the already-signed-in page, compiling each step's route module before any timed test runs. The warming is GET-only, so no welcome marker is stamped and the seeded user stays fresh; the advance PUTs already ride the Hono catch-all warmed by the reset phase. Adds `WARMUP_WELCOME_BUDGET_MS` (45s) as the phase ceiling and raises the setup test timeout 120s → 165s to keep the named-error-before-opaque-timeout invariant. Test-only — no application code changes.

## Test Plan

- `npm run test:e2e` — 37 passed, 0 flaky (3.0m). New warm-up phase ran clean; `welcome.spec` "fresh user completes the welcome flow end to end" passed, confirming the seeded user stays fresh.
- `npm run test:functional` — 510 passed.
- `npx biome check tests/e2e/reset.setup.ts` — clean.
- Local runs hit `next dev`, so this proves no regression + the fresh-user invariant; the cold-start payoff confirms on this PR's CI preview e2e run.

## Checklist

- [x] No `.claude/skills/**/SKILL.md` changes — Skill smoke-test N/A.

Closes #437

### DIFF
--- a/tests/e2e/reset.setup.ts
+++ b/tests/e2e/reset.setup.ts
@@ -17,6 +17,13 @@ setup("reset seeded test users", async ({ baseURL }) => {
 // function spin-up (CI Vercel) cost.
 const WARMUP_ROUTES = ["/signin", "/", "/signup", "/forgot-password"];
 
+// Authed welcome-flow segments. completeWelcome (helpers/session.ts) walks
+// these in every spec that gets past onboarding. The sign-in below only warms
+// the landing (the first step), so the first walk paid the cold profile/
+// programs RSC render and blew the 12s waitForURL on the agreements → profile
+// hop (#437). Warmed GET-only after sign-in — see the phase-3 note.
+const WARMUP_WELCOME_ROUTES = ["/welcome", "/welcome/agreements", "/welcome/profile", "/welcome/programs"];
+
 // Per-attempt navigation windows. The first hit gets a generous window so a
 // slow cold start still has room to respond — a cold function on a degraded CI
 // network (#368) can exceed 30s. The retry is deliberately short: a function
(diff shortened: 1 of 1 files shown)


james-baker:
@NorsemanSpiff this fixes the flake you hit in #433

PR #280, opened 2026-05-25
Buttondown sync: schema, cron, inline resync, bootstrap

james-baker:
## Summary

Wires the app to Buttondown so program membership keeps the newsletter audience and per-program tags in sync. Ships in dry-run by default — `BUTTONDOWN_SYNC_WRITE` gates real PATCHes — so we can land + observe before flipping the switch and retiring the legacy Apps Script.

**Architecture (see `docs/design-buttondown.md`):**
- New `sync_locks` table (lease-based concurrency) + `profiles.buttondown_subscriber_id` + `programs.buttondown_tag` (opt-in per program, NULL = skip). Migration `0010` is pure expand (new table, two nullable columns).
- `src/server/buttondown-sync.ts` — diff-only reconciler: pulls eligible profiles, computes desired tag set per profile, PATCHes Buttondown subscribers only where state diverges.
- `src/server/buttondown-runner.ts` — orchestrates lock acquire/release, retry-on-lock-held with telemetry, Sentry alarms, Axiom structured logging.
- Cron path (broad daily reconcile) + inline path (join/leave/admin-remove + first-profile-save) + admin buttons.
- `scripts/buttondown-bootstrap.ts` — one-shot cutover reconciliation with dry-run and `--write` modes.

**Observability:** HTTP-level Buttondown telemetry, `errorKind`/`durationMs` on every sync, Sentry capture for hard failures, dashboard pointers in `docs/doc-axiom.md`.

**Tests:** 5 new test files (sync, runner, routes, bootstrap, first-save, sync-locks) — full functional suite green.

## Cutover sequence

Merge does **not** start writing to Buttondown. The stage-gated rollout:

1. Merge & deploy with `BUTTONDOWN_SYNC_WRITE` unset — cron + inline run dry-run.
2. Watch a cron or two in Axiom (`message == "buttondown sync"`, `fields.action == "summary"`).
3. `npx tsx scripts/buttondown-bootstrap.ts --prod` — dry-run, eyeball every `PLAN reconcile` line.
4. `npx tsx scripts/buttondown-bootstrap.ts --prod --write` — confirm at prompt.
5. Flip `BUTTONDOWN_SYNC_WRITE=1` in Vercel.
6. Disable the legacy Apps Script.

Before merging: dispatch `prod:db:expand` so migration `0010` is applied to prod (previews share prod DB but skip migrations).

## Deferred follow-ups

- Distinguish cron-vs-inline lock-held for Sentry alerting (mitigated by retry budgets).
[...]

PR #445, opened 2026-06-21
feat(buttondown): sync display name into subscriber metadata.name

### LINKED ISSUE #444: feat(buttondown): sync member display name into subscriber metadata.name
## Context

Buttondown subscribers carry an arbitrary `metadata` object. Earlier iterations of IS Web (manual, then the Apps Script) populated `metadata.name` with the member's display name. When the app took over as the writer (#280, cutover 2026-06-12), that detail was dropped — we sync **tags** but never write `metadata`. Existing subscribers have been manually backfilled, so this is purely about keeping it true going forward.

**Desired value:** `metadata.name` = `profiles.displayName`, **verbatim** as the member supplied it (typically `"Firstname Lastname"`, sometimes first name only). Never the slug, never normalized.

This extends the integration's invariant (`docs/design-buttondown.md` → "The invariant") with: *…and whose `metadata.name` equals their current display name.*

## Why this is more than a one-line add

`metadata.name` should self-heal the same way tags do, which means the daily cron has to **read `metadata` back** to detect drift — and today `projectSubscriber` deliberately drops it. And because these subscribers carry **other** metadata fields we must not clobber, every name write is a **merge** (read-modify-write), not a blind set. Buttondown's PATCH-on-`metadata` semantics (replace-whole-blob vs server-side merge) decide how the merge is implemented and must be confirmed empirically.

## Changes by seam

**1. Client — surface `metadata` (`src/server/buttondown.ts`)**
- Add `metadata?: Record<string, string>` to `ButtondownSubscriber`, `CreateSubscriberInput`, and `UpdateSubscriberInput`.
- Add `metadata` to `projectSubscriber` so reads carry it (the existing comment at `buttondown.ts:122` explains why the projection must stay aligned with the type — this keeps that contract).

**2. Confirm PATCH metadata semantics + re-record golds (`tests/...`, Appendix A flow)**
- Add a probe to `tests/manual/_buttondown-probes.ts` that PATCHes `metadata.name` on a subscriber already holding a second metadata key, and asserts the other key survives. This is the definitive answer to merge-vs-replace and drives the implementation in seam 4/5.
- Teach the in-memory fake (`tests/functional/server/buttondown-fake.ts`) to store/return `metadata` with the **same** create/PATCH semantics the probe reveals.
- Re-record golds: adding `metadata` to the projection changes `typed_result` in **every** gold file, so the replay test fails until `npm run special:buttondown:record-api` is re-run (needs `BUTTONDOWN_TEST_API_KEY`), then `npm run test:manual:buttondown-api-replay` is green. Owner = whoever holds the test-newsletter key.

**3. Set `name` on create (both create paths)**
- `runFirstProfileSaveSync` (`buttondown-sync.ts`): add `displayName` to its profile select; POST-create includes `metadata: { name }`, and the "active, no `isweb-member`" overwrite PATCH merges name into the fetched subscriber's existing metadata.
- `syncOneProfile` catch-up create (cron): `loadEligibleProfiles` selects `displayName`; create includes `metadata: { name }`.

**4. Cron reconciles `name` (`syncOneProfile`)**
- Thread `displayName` into the per-profile desired state. After the tag/email diff, compare `subscriber.metadata?.name` to the desired name; if they differ, add `metadata: { ...subscriber.metadata, name }` to the PATCH (merge preserves other fields). Reuse the existing single-PATCH-per-profile shape; extend the summary/log events with a `name-updated` action.

**5. Name-change push on `PUT /me`**
- In the existing block that already mirrors a `displayName` edit into Supabase auth (`api.ts:448`), add a sibling Buttondown push **after the DB commit**, on the same `parsed.displayName !== undefined` condition.
- Route it through `runProfileResyncForServer` with a new `ProfileResyncReason: "update-name"` rather than a bespoke writer — the merge-preserving name reconciliation from seam 4 then lives in exactly one place, and this mirrors how join/leave already trigger a scoped resync. Same `BUTTONDOWN_SYNC_WRITE === "1"` gate, same swallow-on-failure posture as the auth mirror.

## Decisions / edge cases
- **Null/empty `displayName`:** never blank an existing `metadata.name`; only write when the display name is non-empty. (A saved profile normally has one.)
- **Hidden profiles:** unchanged — still suppressed from *create*, but a name reconcile still applies if a subscriber already exists (consistent with tag handling).
- **Unsubscribed:** no write, as today.

## Non-goals
- No new metadata fields beyond `name`.
- No schema/DB migration — `displayName` already exists; `metadata` lives only in Buttondown. (So no expand-contract step.)
- No new env vars — rides `BUTTONDOWN_API_KEY` + `BUTTONDOWN_SYNC_WRITE`.

## Testing
- Functional: first-save sets `metadata.name`; cron backfills a missing name and repairs a drifted one while preserving a second metadata key; a `PUT /me` name edit pushes; null name doesn't blank.
- Manual: the re-recorded golds + replay (seam 2).
- Update `docs/design-buttondown.md` (invariant, write policy, the "What gets synced" section) and add a `docs/devjournal.md` line.

## Sequencing note
The golds re-record (seam 2) is the long pole: it needs `BUTTONDOWN_TEST_API_KEY` and a person to run it, and the replay test stays red until it's done. If blocking on that is undesirable, the write paths can ship behind the existing dry-run gate and the re-record can land separately.

---

james-baker:
Summary: Mirror each member's display name into their Buttondown subscriber's `metadata.name`, kept true going forward.

Why: Earlier iterations (manual, then the Apps Script) populated `metadata.name`; when the app took over as the writer (#280) that detail was dropped. This restores it, extending the integration's invariant to the display name, not just tags.

Behavior: `metadata.name` (verbatim `profiles.displayName`) is seeded on both create paths, reconciled by the daily cron via a merge-preserving `{ ...existing, name }` PATCH (new `name-updated` action), and pushed inline on a non-first-save `PUT /me` name edit (resync reason `update-name`). An empty/whitespace name never blanks an existing value. The re-recorded golds + new probe 20 pin that Buttondown returns `metadata` as `{}` when unset and replaces the whole blob on PATCH. Two adjacent items rode along: the manual record/seed scripts gain an `allowReservedTestEmails` client opt-out (the reserved-TLD guard had silently broken them against the `@fixture.test` audience), and the dead `409 ButtondownConflictError` branch is removed (duplicates really return `400 email_already_exists`).

Test Plan:
- ran `npm test` locally — functional 565 + 37 e2e green
- re-recorded the Buttondown API golds against the api-tests newsletter; `npm run test:manual:buttondown-api-replay` green

Closes #444

### DIFF
--- a/docs/design-buttondown.md
+++ b/docs/design-buttondown.md
@@ -18,7 +18,7 @@ The app's sync **replaced** the Apps Script as the writer of program tags; the c
 
 One sentence the whole integration enforces:
 
-> A member who has saved their profile should exist as a Buttondown subscriber, tagged exactly according to their current memberships in programs that have a `buttondownTag` set.
+> A member who has saved their profile should exist as a Buttondown subscriber, tagged exactly according to their current memberships in programs that have a `buttondownTag` set, and whose `metadata.name` equals their current display name.
 
 Every code path below is a route to making that statement true.
 
@@ -50,6 +50,17 @@ We identify "tags we manage" by joining against the `programs` table: the set of
 
 Subscribers are keyed by email. We use `auth.users.email` for the matching member's profile id — the same email Supabase signs them in with. There is no separate "newsletter email" field. Edge case: members who already exist in Buttondown under a different email won't be matched; that's an acknowledged limitation.
 
+### Display name (`metadata.name`)
+
+Buttondown subscribers carry an open `metadata` key-value blob. We mirror the member's display name into `metadata.name`, **verbatim** as they supplied it (`profiles.displayName` — typically "Firstname Lastname", sometimes a first name only; never the slug, never normalized). This restores a detail the pre-cutover Apps Script set, which the app initially dropped when it took over as the writer (#444).
+
+Rules:
+
+- **Only ever write a non-empty name.** An empty or whitespace-only display name never blanks an existing `metadata.name`.
+- **Every write is merge-preserving.** Subscribers carry other metadata keys we don't own, so each name write reads the subscriber first and PATCHes `{ ...existing, name }` — a whole-blob write that keeps the rest. Buttondown **replaces** the whole `metadata` blob on PATCH (it does not merge server-side — confirmed by [Appendix A](#appendix-a-fake-vs-real-comparison-recording-from-a-dedicated-newsletter) probe 20), so sending the full merged blob is exactly what preserves the other keys.
+- **Self-heals like tags.** The daily cron reads `metadata.name` back and repairs drift, so `metadata` is part of the client's subscriber projection.
+- **Hidden / unsubscribed:** same posture as tags — no create for hidden profiles, no write at all for unsubscribed subscribers, but a name reconcile still applies to an already-existing subscriber.
+
 ### Standing tags: `isweb-member`, `new`, `returning`
 
 Three tags carry signup-moment semantics — the app writes them when a subscriber first arrives via IS Web, and the daily cron never modifies them after that:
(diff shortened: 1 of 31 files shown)


james-baker:
Running a dry-run on /admin post-merge for the real acid test.

PR #451, opened 2026-06-21
ci(e2e): serialize runs with a concurrency group (#358)

### LINKED ISSUE #358: e2e flake: concurrent runs clobber each other via shared prod DB + seeded accounts
## Summary

The Playwright e2e suite runs against **prod Supabase** (previews share the prod `DATABASE_URL`), authenticating as two fixed seeded accounts (`e2e-regular@testfake.local`, `e2e-admin@testfake.local`). `/api/_test/reset` (`resetE2EUsers` in `src/server/test-reset.ts`) **deletes those accounts' invites and wipes their profile fields**. Nothing prevents two e2e runs from executing at the same time — so when two deploys land within the ~3-minute e2e window, the runs clobber each other's seeded-account state mid-test and fail.

## Observed incident

- main run [27053683966](https://github.com/Intentional-Society/is-app/actions/runs/27053683966) was active 05:23:23 → 05:26:57; the PR #357 branch run [27053701861](https://github.com/Intentional-Society/is-app/actions/runs/27053701861) was created 05:24:18 and still failing at 05:27:33 — ~2.5 min of overlap.
- Both failures were in `tests/e2e/invites.spec.ts`:
  - `valid code → submitting email shows 'check your email'` — the invite note was not visible (the other run's reset `DELETE`d the invite).
  - `create an invite, see it listed, revoke it` — `waitForURL` timeout (the other run's reset blanked the profile, re-triggering the welcome redirect).
- Re-running the failed job in isolation passed immediately → confirms flake, not regression.
- Trigger this time: a rebase force-push re-deployed the PR preview ~1 min after a main deploy. Deploys are normally spaced out, which is why this has stayed latent.

## Scope / safety

Bounded to the two `@testfake.local` test accounts — `resetE2EUsers` filters by `createdBy` / profile `id`, preserves `isAdmin`, and never touches `auth.users` or any real member. **No real member data is at risk**; the impact is e2e flakiness, not data loss.

### Option A — CI concurrency guard (cheap, immediate)

Add a concurrency group to `.github/workflows/e2e.yml` so e2e runs queue instead of overlapping:

```yaml
concurrency:
  group: e2e-shared-prod-db
  cancel-in-progress: false
```

- Pros: ~3 lines, no app changes, fully closes the overlap gap. `cancel-in-progress: false` queues runs so no deploy loses its e2e.
- Cons: serializes e2e globally — when several deploys land together they run one at a time (slower feedback, though each still runs).

### Option B — per-run isolated test accounts (robust, larger)

Give each e2e run its own unique accounts (e.g. emails suffixed by run id / a short random token) instead of the two shared seeded ones, so concurrent runs can't collide and e2e can stay parallel.

- Pros: true isolation; removes the shared-state assumption entirely; keeps runs parallel.
- Cons: needs per-run account provisioning + teardown (a service-role key in CI, or a create/cleanup flow), `resetE2EUsers` rescoping, and password/secret handling. Materially more work.

## Recommendation

Ship Option A now as the guard; revisit Option B if/when e2e parallelism or volume makes serialization painful.

---

oolu4236:
## Summary

The Playwright e2e suite runs against the **shared prod Supabase DB** and authenticates as two fixed seeded accounts, whose state `/api/_test/reset` wipes mid-run. Nothing stopped two e2e runs from executing at once, so when two deploys landed within the ~3-min e2e window they clobbered each other's seeded-account state and failed (the recurring `invites.spec.ts` flake we've hit repeatedly).

Adds a global **concurrency group** to `e2e.yml` so runs queue instead of overlapping:

```yaml
concurrency:
  group: e2e-shared-prod-db
  cancel-in-progress: false
```

`cancel-in-progress: false` queues runs rather than cancelling, so no deploy loses its e2e result. This is **Option A** from the issue; per-run isolated accounts (Option B) is deferred until e2e volume makes serialization painful.

## Test plan

- [x] `e2e.yml` parses as valid YAML; `concurrency` is a top-level key.
- [ ] CI lint + functional pass.
- [ ] After merge, overlapping deploys queue their e2e runs instead of running concurrently (observable on the Actions tab).

Note: the concurrency rule takes effect once merged to `main` (deployment_status workflows resolve from the default branch).

Closes #358

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -3,6 +3,15 @@ name: E2E Tests
 on:
   deployment_status:
 
+# Serialize e2e across all branches: the suite runs against the shared prod
+# Supabase DB and drives two fixed seeded accounts, resetting their state
+# mid-run. Two overlapping runs clobber each other's accounts and fail (#358).
+# A single global group queues runs instead; cancel-in-progress: false so no
+# deploy loses its e2e result.
+concurrency:
+  group: e2e-shared-prod-db
+  cancel-in-progress: false
+
 # See ci.yml for the rationale. upload-artifact@v4 works with contents:
 # read — it doesn't need write scope on any top-level resource.
 permissions:
--- a/docs/devjournal.md
+++ b/docs/devjournal.md
@@ -8,6 +8,10 @@ Each entry: **Date** | **Author** | **Title**, followed by description text. Mos
 
 Avatar signed URLs rotate a `?token=` that's part of next/image's optimizer cache key, and the optimizer's default `minimumCacheTTL` is 4h — so every avatar's 1024² source was re-fetched from Storage daily (rotation) and again every 4h (expiry), blowing the free-plan egress quota during the Convening. Fix: `SIGN_TTL_SECONDS` 24h → 5 days, and `images.minimumCacheTTL` → 31 days (safe — object paths are immutable). The durable fix (a tokenless proxy route) stays open as #382.
 
+## 2026-06-21 | Ola | e2e runs serialize via a CI concurrency group
+
+Two e2e runs overlapping against the shared prod DB clobbered the seeded accounts mid-run and failed. A global `concurrency: { group: e2e-shared-prod-db, cancel-in-progress: false }` in `e2e.yml` now queues runs instead — no deploy loses its e2e. (#358)
+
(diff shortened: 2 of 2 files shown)


james-baker (approved):
workflow looks good, thanks.

Not a big deal: I always tell the AIs to write only half as much (as they always do despite "be terse" memories) in the devjournal, it could be shortened when you're resolving the merge conflict.

benjifriedman (approved):

PR #454, opened 2026-06-23
feat(admin): permanently delete a member's account (#185)

### LINKED ISSUE #185: No way to delete your account

oolu4236:
## Summary

Implements the **admin-only account deletion** half of #185 (the member-facing **deactivate** half shipped in #301). Per James's 2026-05-27 decision: members self-serve via deactivate; permanent deletion is an admin operation.

An admin deletes a member from `/admin/members`:
- **Server** — `deleteMemberAccount(targetId, requesterId)` (`members-admin.ts`): `clearAvatar` → delete the `profiles` row (FK cascade clears program memberships, relations, and invite hints; null-sets `invites.createdBy`/`redeemedBy` so the chain survives anonymized) → `supabaseAdmin.auth.admin.deleteUser`. Order matters — `profiles → auth.users` is `ON DELETE no action`, so the profile (child) goes first; the single-statement delete avoids a pooler transaction.
- **API** — `DELETE /admin/members/:id` in the `requireAdmin` sub-router → `self_delete`→403, `is_admin`→409, `not_found`→404.
- **UI** — per-row "Delete account" → a **type-to-confirm** panel (type the member's display name to enable), listing what's removed (profile, avatar, relationships, memberships) vs preserved (invites sent/redeemed, anonymized).
- **Guards** (mirror `setAdminStatus`): refuse self-delete (use deactivate), refuse deleting a fellow admin (demote first).

## Test plan

- [x] New `admin-members-delete.test.ts` (7 cases): deletes a member (profile + memberships gone, `deleteUser` called); cascades relations; anonymizes the member's invites (`createdBy` → null); self-delete → 403; admin → 409; unknown → 404; non-UUID → 400; non-admin caller → 404.
- [x] Full functional suite green (583 tests); lint + typecheck clean.
- [ ] Manual: as admin on `/admin/members`, delete a throwaway member via type-to-confirm; verify they vanish from the directory + DB; verify self/admin guards block.
- [ ] CI lint + functional + e2e pass.

## Out of scope (follow-up)
- GDPR "right to be forgotten" as a distinct manual process (James's open question) — this hard delete is v1; a documented manual path can follow.
- No soft-delete/grace period (issue says hard delete is acceptable for v1).

Closes #185

### DIFF
--- a/src/app/admin/members/members-admin-panel.tsx
+++ b/src/app/admin/members/members-admin-panel.tsx
@@ -1,10 +1,12 @@
 "use client";
 
-import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
+import { type UseMutationResult, useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
 import { ShieldCheck } from "lucide-react";
+import { useState } from "react";
 
 import { Avatar } from "@/components/avatar";
 import { Button } from "@/components/ui/button";
+import { Input } from "@/components/ui/input";
 import { apiClient } from "@/lib/api";
 import type { AdminMember } from "@/lib/api-types";
 
@@ -13,6 +15,11 @@ const STALE_TIME = 5 * 60 * 1000;
 
 type ToggleVars = { member: AdminMember; isAdmin: boolean };
 type ToggleError = Error & { memberId?: string };
+type ToggleMutation = UseMutationResult<ToggleVars, ToggleError, ToggleVars>;
+
+type DeleteVars = { id: string };
+type DeleteError = Error & { memberId?: string };
+type DeleteMutation = UseMutationResult<DeleteVars, DeleteError, DeleteVars>;
 
 const fetchMembers = async (): Promise<AdminMember[]> => {
   const res = await apiClient.api.admin.members.$get();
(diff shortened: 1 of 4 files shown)


james-baker:
hey @oolu4236 was the "check constraint" failure bug addressed in some way?

oolu4236:
> hey @oolu4236 was the "check constraint" failure bug addressed in some way?

 it wasn't addressed; that path aborted on `invites_redemption_pair` for any invite-redeemed member. Fixed in #464 (clears the redemption pair before the profile delete) with a regression test that fails without the fix. Thanks for flagging.

Issue #463, opened 2026-06-24
/ship merge gate: harness ask verified firing in all four permission modes (2026-07-07); #460 cause unconfirmed — Y/n restore recipe inside


NorsemanSpiff:
## Problem
`/ship` confirms every production merge via a checked-in `.claude/settings.json` `ask` rule on
`gh pr merge` (Bash + PowerShell). #459 deleted `/ship`'s conversational `Merge? (y/n)` on the premise
that this harness `ask` is the sole per-merge confirmation. **A controlled test (2026-06-26) confirms
the `ask` *does* fire per-merge in default mode** — so the gate is far more reliable than #460 alone
implied. **Severity: LOW.**

## Decision (2026-06-26, Blake): keep #459 — do NOT restore the Y/n for now
The two-merge same-session test refuted the "gate is broken" framing, and #460 is most likely `auto`
mode (see Behaviors). So we keep the current state: the checked-in `gh pr merge` `ask` rule is
`/ship`'s merge confirmation, with **no** conversational Y/n. The prepped revert branch
`restore-ship-merge-yn` will be **deleted** rather than shipped. **If we choose to add the Y/n back
later, the precise recipe is in "How to restore…" below** so the deleted branch isn't needed.

## `/ship` invocation flag (scope)
`/ship` keeps `disable-model-invocation: true` — it fires only on an explicit, human-typed `/ship`
(`/commit` and `/pr` are NL-invocable; `/ship` isn't). So this merge path is only ever entered by a
human, which bounds the blast radius.

## Behaviors
- **#460 merged silently (2026-06-24)** — a `/ship` merge ran `gh pr merge` and merged to `main` with
  **no prompt**, both settings files correct.
- **#461 (fresh session) prompted** — a fresh session's first merge prompted (declined).
- **Two-merge same-session test (2026-06-26) → REFUTES the suppression hypothesis.** In **default
  mode**: PR #465 approved with plain "Yes", then PR #466 shipped in the **same session** → **#466
  still prompted.** Ordinary approval does **not** disarm the gate; `ask` fires per-merge. The
  "once-per-session suppression after the first approval" claim is **disproven.**
- **Leading explanation for #460: the #459/#460 session was in `auto` mode** (auto-approves tool
  calls) — so those merges were auto-approved, not gated. Fits the docs (auto mode auto-approves with
  safety checks; unlike `bypassPermissions`, it is **not** documented to honor `ask` rules), explains
[...]

PR #467, opened 2026-06-26
docs(skills): note /ship merge gate isn't absolute in auto mode (#463)

NorsemanSpiff:
Corrects an overclaim in shipped artifacts and completes the Thread-15 / #463 decision (*keep #459 + document the caveat*).

## Why
#459 deleted `/ship`'s conversational Y/n on the premise that the checked-in `gh pr merge` `ask` rule is the **sole, durable** merge confirmation. A two-merge same-session test (2026-06-26) showed the `ask` **does** fire per-merge in **default** mode — but it is **not an absolute gate**: `auto` mode auto-approves it, and "don't ask again" silences it for the session (that's how #460 merged silently, in an auto-mode session). Severity is LOW (the default-mode gate works), but the shipped docs still said "sole/durable" — this fixes that.

## What (docs/instruction wording only — no behavior change)
- `.claude/skills/ship/SKILL.md` step 10/11 — the `ask` is the confirmation **under default mode**, not absolute; auto-mode / "don't ask again" caveat + a pointer to #463.
- `docs/plan-skill-nl-invocation.md` — a dated ⚠️ Update note (Thread 15 resolved).
- `docs/spec-portable-ai-procedures.md` — a step-10 caveat + a one-line note in the B.1 walkthrough.

The precise (unconditional) Y/n-restore recipe is preserved in #463 if ever revisited.

## Test Plan
- `npm run lint`: clean.
- No app/test code touched.

Refs #463.


NorsemanSpiff:
Fixed in 8a708dd — scoped step 10's lead to "the merge confirmation **under default permission mode** in every case", matching the parallel `ship/SKILL.md` step 10 and the caveat. Good catch on the self-contradiction.

Issue #287, opened 2026-05-26
fix: add stash hygiene guidance and cross-branch stash detection to prevent unexpected merge conflicts


NorsemanSpiff:
> **Edited 2026-05-26 (NorsemanSpiff):** Reframed root cause and acceptance criteria after working through the issue with Claude. The original Problem Statement and Options A/B/C are preserved in the "Previous framing" section at the bottom for audit trail. See companion note on #284.

## Problem Statement

The 2026-05-25 incident (RCA #284) was *triggered* by a specific shell antipattern, not by stash hygiene in general.

The idiom `git stash && <command>; git stash pop` was used as a verification step against an apparently-clean tree. On a clean tree, **`git stash` is a no-op** — it does not push an "empty marker" onto the stash stack. The trailing `git stash pop` therefore applies whatever entry was already on top of the stack (in this case, leftover work from `env-example-preflight`), producing a surprise conflict in `scripts/check-env.mjs`.

Cross-branch stash accumulation is the background condition that made the hazard reachable, but the direct cause was using a side-effecting stash dance as a read-only verification step. The correct tool for "is there anything that would interfere?" is `git stash list`, which has no side effects.

## Recommended Fix — Option D (narrow)

Add a short callout to `docs/strategy-committing.md` under "Before committing" that names the antipattern and the safe alternative. No tooling. No preflight check. No new doc file.

Draft text (subject to wordsmithing in the PR):

> **Don't use `git stash && X; git stash pop` to verify something against a clean tree.** `git stash` is a no-op when there's nothing to stash, so the following `git stash pop` applies whatever was already on top of the stash stack — typically leftover work from another branch, which then conflicts. To inspect pending stashes, use `git stash list` (read-only).

Consistent with @james-baker's "first, do no harm" framing on #284: this is the minimum-viable workflow correction without introducing tooling or process the team doesn't need.

## Acceptance Criteria

- `docs/strategy-committing.md` includes a brief verification-gotcha note calling out the `git stash && X; git stash pop` antipattern and recommending `git stash list` for read-only inspection
- `docs/devjournal.md` entry records the correction and links #284 and this issue

## Priority

Short-term, low effort (one paragraph of docs).

## Related

- Parent RCA: #284 (see "Note on Gap 4 root cause" comment for the refinement)
- Related: #288 (node_modules staleness check — separate preflight concern, not affected by this reframing)

---

<details>
[...]

Issue #286, opened 2026-05-26
chore: run biome migrate to fix biome.json schema version pin (2.4.14 → 2.4.15)


NorsemanSpiff:
## Problem Statement

`biome.json` pins the `$schema` URL to a specific Biome version (`2.4.14`), but the installed `@biomejs/biome` package is `2.4.15`. Every `npm run lint` run produces an info-level advisory:

```
biome.json:2:14 deserialize ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  i The configuration schema version does not match the CLI version 2.4.15

    1 │ {
  > 2 │   "$schema": "https://biomejs.dev/schemas/2.4.14/schema.json",
      │              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    3 │   "vcs": {
  i   Expected: 2.4.15
      Found:    2.4.14
  i Run the command biome migrate to migrate the configuration file.

Checked 201 files in 172ms. No fixes applied.
Found 1 info.
```

While this is info-only (no lint errors, no blocked CI), the advisory fires on every lint run and will recur after any future Biome version bump that isn't followed by `biome migrate`.

**Observed in RCA #284 (2026-05-25):** Biome `2.4.14 → 2.4.15` bump arrived via a dependency update; `biome migrate` was not run. The advisory does not block the PR or CI (docs-only path filter skips lint on that PR), but it is persistent noise on all developer machines.

## Scope

- `biome.json` — update `$schema` URL via `biome migrate`
- Consider switching to a `latest` schema URL to avoid future drift (see decision below)

## Proposed Implementation Approach

**Step 1 — Run `biome migrate`:**
```
npx biome migrate
```
This updates `biome.json`'s `$schema` to match the installed version and applies any non-breaking config migrations introduced between 2.4.14 and 2.4.15. Review the diff before committing.

**Step 2 — Decide on schema URL strategy:**

| Option | Pros | Cons |
|---|---|---|
| Keep pinned URL (update on each bump) | Explicit; schema validated against exact version | Requires `biome migrate` after every bump; easy to forget |
| Switch to `latest` URL | Never drifts; zero maintenance | `latest` schema may not exist as a stable URL in Biome's CDN (verify before adopting) |
| Keep pinned URL + add to update workflow | Best of both: explicit + automated | Requires process discipline |

[...]

PR #476, opened 2026-06-28
fix(dev): catch stale node_modules before local tooling

### LINKED ISSUE #284: RCA: Local dev environment gaps surfaced during Step 6 (missing Playwright binary, biome schema pin, dangling stash)
## Executive Summary

During Step 6 of the PR #133 workflow (opening the dev-plan PR on 2026-05-25), three pre-existing local environment gaps were surfaced when running `npm test`, and a fourth systemic gap was identified through root-cause analysis. None blocked CI or the PR. All four are reproducible failure modes that any developer can hit independently of their code changes.

## Impact

- **Scope:** Local development only. CI (`Lint & Functional Tests`, CodeQL, Vercel) was unaffected on PR #283.
- **Functional tests:** 272/272 passed — application code is correct.
- **PR #283:** Opened successfully; all CI checks green.
- **Risk if unaddressed:** New contributors will hit cryptic binary-not-found errors; future devDependency additions will silently break local test runs; dangling stashes will continue to cause unexpected merge conflicts.

## Detection

Manual `npm test` run during pre-commit validation on branch `plan/portable-ai-procedures`, 2026-05-25.

### Symptom 1 — `'biome' is not recognized`
`npm test` failed immediately: `biome` binary not in `node_modules/.bin/`. `npm install` had not been run in this session.

**Root cause:** No preflight step checks whether `node_modules` is current with `package-lock.json`. See #288 for the systemic fix.

### Symptom 2 — Biome schema version mismatch
After `npm install`, lint produced a schema advisory: `biome.json` pins schema `2.4.14` but installed Biome is `2.4.15`. Info-only; no lint errors. Biome checked 201 files successfully.

**Root cause:** `biome migrate` was not run after Biome was bumped from 2.4.14 → 2.4.15. No automation triggers it. See #286.

### Symptom 3 — All 25 Playwright e2e tests failed
Error: `Executable doesn't exist at ...chromium_headless_shell-1223/chrome-headless-shell.exe`. Test logic was never reached; the browser binary was simply missing.

**Root cause:** `npx playwright install` is not part of `npm run setup` or documented in `docs/setup-dev-machine.md`. See #285.

### Symptom 4 — Cross-branch stash conflict (systemic)
`git stash pop` accidentally applied a stash entry from an unrelated branch (`env-example-preflight`), causing a merge conflict in `scripts/check-env.mjs`. Conflict was resolved cleanly; no changes were lost.

**Root cause:** No guidance or tooling prevents developers from accumulating stash entries across branches. A blind `git stash pop` applies the top-of-stack entry regardless of which branch it originated from. See #287.

## 5 Whys

**Gap 1 — node_modules not installed:**
1. Why did `biome` fail? → binary not present in `node_modules/.bin/`
2. Why not present? → `npm install` not run in this session
3. Why wasn't there a clear early failure? → `npm test` starts with `npm run lint`, not a dependency check
4. Why no dependency check? → No preflight step compares `package-lock.json` mtime against `node_modules/.package-lock.json`
5. Why not? → The April 2026 preflight improvements targeted `.env.local` drift; node_modules staleness was never addressed → **Fix: #288**

**Gap 2 — Biome schema pin drift:**
1. Why the advisory? → `biome.json` `$schema` references 2.4.14; installed version is 2.4.15
2. Why the mismatch? → `npm install` updated the binary; `biome.json` was not updated
3. Why not? → `biome migrate` not run after the version bump
4. Why not? → No automation or checklist item triggers it on version bumps
5. Why not? → Never added to the update workflow → **Fix: #286**

**Gap 3 — Playwright browser binary missing:**
1. Why did e2e tests fail? → Chromium binary not present
2. Why not present? → `npx playwright install` not run
3. Why not run? → Not documented in `docs/setup-dev-machine.md`
4. Why not documented? → Never added when Playwright was introduced
5. Why not? → No process requires updating setup docs when a test infrastructure dep is added → **Fix: #285**

**Gap 4 — Cross-branch stash conflict:**
1. Why did the conflict occur? → `git stash pop` applied a stash from a different branch
2. Why was that stash present? → Work on `env-example-preflight` was stashed and never committed or dropped
3. Why not cleaned up? → No stash hygiene guidance or tooling
4. Why not? → Stash lifecycle is not part of the branching workflow docs
5. Why not? → Never added → **Fix: #287**

## Contributing Factors

- `npm run setup` has only one step (create `.env.local`); it has never been extended to cover node_modules or browser binaries
- `biome.json` uses a pinned `$schema` URL rather than a version-agnostic reference
- No process gate requires updating `docs/setup-dev-machine.md` when new test infrastructure deps are added

## Timeline

| Time (approx) | Event | Result |
|---|---|---|
| ~17:22 | `npm test` (first run, no prior `npm install`) | FAIL — `'biome' is not recognized` |
| ~17:23 | `git stash && npm test \| head -20; git stash pop` (verification attempt) | BLOCKED — stash pop conflicted with `scripts/check-env.mjs` from `env-example-preflight` stash |
| ~17:23 | `git restore --staged scripts/check-env.mjs && git restore scripts/check-env.mjs && git stash drop stash@{0}` | Recovery — clean working tree restored |
| ~17:24 | `npm install` | Biome binary now installed |
| ~17:25 | `npm test` (second run) | PARTIAL — lint: 201 files, 0 errors, 1 info (schema pin); typecheck: pass; dev:db: pass; vitest: pass; Playwright: 25 tests FAIL (browser binary missing) |
| ~17:27 | `npm run test:functional` (isolated) | PASS — 17 files, 272 tests |
| ~17:27 | RCA mode entered | Investigation begins |
| ~17:35 | Branch rebased onto current `origin/main`; committed; pushed | PR #283 opened |
| CI | `Lint & Functional Tests`: pass (5s, docs-only path filter); CodeQL: pass; Vercel: ignored | All checks green |

## Environment Snapshot

| Item | Value |
|---|---|
| OS | Windows 11 Pro 10.0.26200 |
| Node.js | v22.22.2 |
| npm | 10.9.7 |
| Playwright | @playwright/test@1.60.0 |
| Playwright browser installed? | No — `chromium_headless_shell-1223` missing |
| Biome installed (after npm install) | 2.4.15 |
| Biome schema pin | 2.4.14 (`biome.json`) |
| Biome introduced | 2026-05-07 (`Replace ESLint with Biome`, PR #134) |
| April preflight improvements | 2026-04-23–28 (env-key drift only; node_modules not covered) |

## What Went Well

- Isolated functional test run (`npm run test:functional`) correctly scoped the failure — confirmed app logic is correct and test infrastructure is the issue
- Stash conflict resolved cleanly with no changes lost
- CI path filter for docs-only diffs worked as designed — required check green in 5s

## What Slowed Diagnosis

- Initial `git fetch origin main` sent to `FETCH_HEAD` rather than updating `origin/main`, causing a 34-commit stale branch (resolved by `git fetch origin && git rebase origin/main`)
- Verification attempt using `git stash && ...pop` was the wrong tool — `git stash list` inspection is the correct way to check for pre-existing state without side effects

## Action Items

| Issue | Title | Priority |
|---|---|---|
| #285 | Add `npx playwright install` to setup docs and `npm run setup` | Short-term |
| #286 | Run `biome migrate` to fix biome.json schema version pin (2.4.14 → 2.4.15) | Medium-term |
| #287 | Add stash hygiene guidance and cross-branch stash detection | Short-term (docs); Medium-term (tooling) |
| #288 | Add node_modules staleness check to dev:db preflight | Short-term |

## Risk Assessment

- **Low** — all gaps are local environment only; no data loss; no security impact; CI unaffected
- **Medium** recurrence risk for new contributors without this RCA context

---

## Appendix A: Hypotheses Tracked During Investigation

| # | Hypothesis | Final Status | Confidence |
|---|---|---|---|
| H1 | `npm test` fails because `biome` binary not installed | Confirmed | 1.0 |
| H2 | After `npm install`, lint fails with a hard error blocking commit | Weakened — info-only advisory, not a hard error | 0.2 |
| H3 | E2E failures are browser-binary-missing, not test logic | Confirmed | 1.0 |
| H4 | Functional tests pass independently | Confirmed | 1.0 |
| H5 | Biome schema mismatch is pre-existing (not caused by this PR) | Confirmed | 1.0 |
| H6 | E2E failures are pre-existing (not caused by this PR) | Confirmed | 1.0 |

## Appendix B: CI Path Filter Behavior

The `Lint & Functional Tests` workflow uses `dorny/paths-filter` with `predicate-quantifier: 'every'` and this filter:

```yaml
code:
  - '**'
  - '!docs/**'
  - '!CLAUDE.md'
```

For PR #283 (only `docs/plan-portable-ai-procedures.md` changed), `filter.outputs.code == 'false'` — lint and functional test steps are skipped entirely. The required check still posts a green status in ~5s. This is the designed behavior for docs-only PRs and is why local lint/test failures did not surface on CI for this PR.

## Related

- PR #283: https://github.com/Intentional-Society/is-app/pull/283 (the PR during which this was surfaced)
- PR #133: https://github.com/Intentional-Society/is-app/pull/133 (parent spec PR for context)

### LINKED ISSUE #288: fix: add node_modules staleness check to dev:db preflight (catches new devDependency binaries before test failures)
## Problem Statement

When a new `devDependency` is added to `package.json` (e.g., a linter, codegen tool, or test runner), developers who already have a `node_modules` directory from before that change will see a cryptic binary-not-found error the next time they run `npm test` or `npm run dev`. The current preflight chain (`check-env.mjs` → `ensure-docker.mjs` → `ensure-supabase.mjs`) has no step that detects whether `node_modules` is current with `package-lock.json`.

**Observed in RCA #284 (2026-05-25):** `@biomejs/biome` was added as a devDependency on 2026-05-07. A developer session on 2026-05-25 ran `npm test` without first running `npm install`, producing `'biome' is not recognized as an internal or external command`. The error appeared at the `npm run lint` step — not at the beginning of the chain where a preflight check would catch it early.

## Why This Gap Exists

The April 2026 preflight improvements (PR #90, `d61d98c`–`83f2eb2`) correctly targeted `.env.local` key drift: they check that your local environment file has all the keys declared in `.env.local.example`. That pattern works well for env var drift. But it does not extend to binary/dependency drift.

Every new `devDependency` added after a developer last ran `npm install` creates the same gap. Without a staleness check, the failure surface is wherever the binary is first invoked — which could be anywhere in the test or build chain, producing an error that looks like a test failure rather than a missing-install failure.

## Root Cause

`npm install` writes a copy of the resolved lockfile to `node_modules/.package-lock.json` (introduced in npm 7). If `package-lock.json` has been modified since the last install, `node_modules/.package-lock.json` will be older. This is a reliable, fast, filesystem-local signal that `npm install` needs to be re-run — and no preflight step checks it.

## Scope

- `scripts/ensure-deps.mjs` (new) — or add a function to `scripts/setup.mjs` / `scripts/ensure-docker.mjs`
- `package.json` `dev:db` script — add `node scripts/ensure-deps.mjs &&` as the first step
- `docs/setup-dev-machine.md` — document the check and what it does

## Proposed Implementation Approach

Add a new preflight script `scripts/ensure-deps.mjs`:

```js
#!/usr/bin/env node
// Preflight: verify node_modules is current with package-lock.json.
// Compares mtime of node_modules/.package-lock.json (written by npm install)
// against package-lock.json. If package-lock.json is newer, dependencies
// may be stale — prompt the developer to run npm install.

import { statSync, existsSync } from "node:fs";
import { resolve } from "node:path";

const root = resolve(import.meta.dirname, "..");
const lockfile = resolve(root, "package-lock.json");
const installedLock = resolve(root, "node_modules", ".package-lock.json");

if (!existsSync(installedLock)) {
  console.error("ensure-deps: node_modules not found or incomplete.");
  console.error("  Fix: run `npm install`.");
  process.exit(1);
}

const lockMtime = statSync(lockfile).mtimeMs;
const installedMtime = statSync(installedLock).mtimeMs;

if (lockMtime > installedMtime) {
  console.error("ensure-deps: package-lock.json is newer than node_modules.");
  console.error("  New dependencies may have been added since you last ran npm install.");
  console.error("  Fix: run `npm install`.");
  process.exit(1);
}

console.log("ensure-deps: node_modules is current.");
```

Then update `dev:db` in `package.json`:
```
"dev:db": "node scripts/ensure-deps.mjs && node scripts/check-env.mjs && ..."
```

**Alternative approach:** Instead of mtime comparison, diff the contents of `package-lock.json` and `node_modules/.package-lock.json` directly (more robust against clock skew, but slower on large lockfiles). The mtime approach is fast and sufficient for this use case.

**Note on CI:** CI runs `npm ci` (or equivalent) which always installs from scratch. This check would never fire in CI. It is a developer-workflow-only guard.

## Acceptance Criteria

- Running `npm test` or `npm run dev` on a repo where `npm install` hasn't been run since a new `devDependency` was added produces a clear, actionable error at the top of the chain — not a cryptic binary-not-found error deep in lint or test output
- The check exits 0 (no-op) when `node_modules` is current — zero overhead for developers who ran `npm install`
- The check exits 1 with an actionable error message when `package-lock.json` is newer than `node_modules/.package-lock.json`
- CI is unaffected
- `docs/setup-dev-machine.md` documents the check

## Test/Verification Plan

1. Run `npm install` (baseline)
2. Add a dummy `devDependency` to `package.json` and update `package-lock.json` (or just touch `package-lock.json` to advance its mtime)
3. Run `npm run dev:db` — confirm the new check fires with an actionable error before any other preflight step
4. Run `npm install`
5. Run `npm run dev:db` again — confirm the check passes cleanly

## Priority

Short-term — this gap will recur with every new devDependency added; the fix is low-risk and fast

## Related

Parent RCA: #284
Related: PR #133 (portable AI procedure framework) — the `/commit` Skill being built as part of that spec could also surface this warning during its pre-flight phase as a complementary (not replacement) check

NorsemanSpiff:
Summary: Add a dependency freshness preflight before local dev and lint entry points.

Why: Developers can pull a package-lock.json change while keeping an older node_modules tree, causing cryptic binary-not-found failures when tools like Biome run.

Behavior: Local dev/test commands now compare package-lock.json with npm's installed node_modules/.package-lock.json and fail early with npm install guidance when dependencies are stale.

Issue alignment / deviations:
- Completes parent RCA #284 by closing its final remaining action item; #285, #286, and #287 are already closed.
- Matches #288's proposed mtime-based check using npm's installed lockfile at node_modules/.package-lock.json.
- Matches #288's dev:db preflight goal so npm run dev, test:functional, test:e2e, and watch fail early when dependencies are stale.
- Intentional deviation: also wires the check before lint because npm test invokes Biome before dev:db; this is necessary to catch the original missing-Biome failure mode at the top of npm test.
- Intentional deviation: leaves `scripts/setup.mjs` untouched because this bug is about detecting stale dependencies at dev/test runtime. After #285, `setup.mjs` owns post-install setup gaps like Playwright browsers; `ensure-deps` owns detecting when `npm install` itself needs to be rerun.

Test Plan:
- ran `node scripts/ensure-deps.mjs`
- ran `npx vitest run --project functional-server tests/functional/server/ensure-deps.test.ts`
- `tests/functional/server/ensure-deps.test.ts` covers #288's manual stale-lockfile test plan with temp repo fixtures: current install passes, missing installed lockfile fails, package-lock newer than node_modules/.package-lock.json fails, and npm test wiring reaches ensure-deps before Biome.
- ran `npm test` locally

Closes #288
Closes #284

Note: AI co-author identity provided by human; auto-detection failed.

Co-Authored-By: GPT-5 Codex <noreply@unspecified>


NorsemanSpiff:
Reviewed this offline with Claude last night once my weekly rate limit reset, going through the Codex implementation in detail. Overall it's a clean, well-tested fix and the wiring decisions are sound — including the call to also gate `lint` (not just `dev:db` as #288 specified), which is what actually catches the original #284 failure since `npm test` runs `lint` first. Nothing here is blocking; approving. A few non-blocking notes and suggestions for follow-up:

**1. mtime comparison can over-report under our own workflows.** The mtime-vs-content tradeoff from #288 is the right default for speed, but two things specific to this repo raise the false-positive rate: trunk-based dev with frequent `git switch`, and worktree lanes reusing slots via `git switch`. Switching A→B→A rewrites `package-lock.json`'s mtime to *now*, so the preflight reports "stale" even when A's deps are already installed — prompting an unnecessary `npm install`. The failure direction is the safe one (it over-reports and self-corrects), so I'd ship the mtime check as-is. Two suggestions:

- **(a) Do now:** add a one-line comment in `ensure-deps.mjs` noting mtime can over-report under branch switching, so the tradeoff is explicit for the next maintainer. Cheap, no risk.
- **(b) Defer — only if this actually bites:** *if* branch-switch false positives become a real annoyance in practice, the upgrade path is a content comparison instead of mtime. Worth flagging the cost so nobody picks this up prematurely: it's **not** a byte-equality check — `node_modules/.package-lock.json` is derived from the installed tree and isn't identical to `package-lock.json`, so a correct comparison has to normalize to npm's hidden-lockfile shape (per-package `version`/`resolved`/`integrity`), which couples us to npm internals. Perf isn't the driver (~50ms on our 600KB lockfile, lost in the noise next to Docker/Supabase boot). So: **no action now** — just the documented fallback if the mtime annoyance shows up.

**2. Preflight runs twice per `npm test`** — once via `lint`, once via `dev:db`. It's two `stat` calls so the cost is nil; flagging only because the gate now lives in two places rather than one.

**3. `format`/`typecheck`/`build` are unguarded.** Same cryptic-binary failure surface; `format` in particular is a plausible first-command-after-pull. Scoping to `npm test` + `dev` is defensible, so I'd leave it — but worth a deliberate decision rather than an accident.

**4. Minor test nit:** the wiring assertions read `package.json` via a CWD-relative path (`readFileSync("package.json", …)`). Works because Vitest runs from repo root, but the temp-repo tests already use absolute paths — `import.meta.dirname` would make it symmetric and CWD-independent.

None of these block merge.

---

NorsemanSpiff:
Circled back to close out the notes above:
1. mtime over-report — added the safe-direction comment; content-diff stays deferred because `node_modules/.package-lock.json` is not byte-identical to `package-lock.json`.
2. Double-run in `npm test` — kept as-is; the preflight is cheap and preserving composable script guards is clearer than adding dedupe state.
3. Unguarded `format`/etc. — added the preflight to `format` so direct Biome writes hit the early dependency check instead of failing with a cryptic tooling error.
4. Test CWD path — switched package script assertions to resolve `package.json` from `import.meta.url` using the existing test convention.

`npm test` green. Nothing else blocking locally.

---

james-baker (approved):
Thanks, love it!

PR #484, opened 2026-07-02
fix(skills): announce at the routing decision + narrate delegation

### LINKED ISSUE #353: Allow 0-to-3 skills to be model-invoked

NorsemanSpiff:
Summary: Relocate the `Using /commit` / `Using /pr` announcement from inside Step 0 to the routing decision, make each parent announce its delegation handoff, and rewrite the over-trigger evals to semantic cases. Follow-on ("PR3") to the merged #353 natural-language-invocation work.

Why: In a cold verification run the announcement showed up reliably only 5 times of 6 — the miss called the Skill tool first and then skipped the in-Skill (Step 0) announce. The Skill body only loads after the tool fires, so the instruction to announce has to live where the model actually decides to route: CLAUDE.md and the frontmatter description. Delegation is the worst case (a delegated child is always invoked tool-first) and wasn't exercised in that run, so the /ship → /pr → /commit chain could run with no visible handoff.

Behavior:
- CLAUDE.md and each /commit and /pr description now say to announce `Using /<skill>` as you route — the first line of the message that calls the Skill tool, before the call. The description budget moved 300 → 350 to fit the clause (both stay ≤350; /ship is unaffected — its description isn't always loaded).
- Step 0's announcement is now a backstop ("if you haven't already announced, do so now"), so the line prints exactly once.
- Each parent announces its handoff at its delegation step (/pr step 4, /ship step 3): it prints `Using /commit — delegated from /pr` (or `Using /pr — delegated from /ship`) as it writes the marker and invokes the child. The child stays quiet when the delegation marker is present — the marker now covers both the Step-0 confirmation and the self-announce.
- evals: commit-7 rewritten to a semantic over-trigger (a bare "yes" to an "explain /commit" offer must not fire); new commit-8 ("commit" as a question topic) and ship-5 ("remind me what /ship does" explains, not ships); new ship-6 (the full cascade announces each hop once); pr-3 extended (delegated /commit narration); commit-5(b) fixed. Counts: commit 8 / pr 9 / ship 6.
- docs: plan-skill-nl-invocation.md reorganized around a PR1/PR2/PR3 structure and updated with the cold re-verify results; spec §2 updated.

Test Plan:
- ran `npm test` locally — full suite green; Playwright e2e 37/37; lint + typecheck + Vitest passed ahead of it.
- `npx biome check evals/evals.json`: clean; evals.json valid JSON; counts 8 / 9 / 6.
- Cold re-verify in fresh sessions (2026-07-01): direct affirmation announced `Using /commit` at the routing decision 3/3, exactly once, routed through the Skill, Step 0 → Stop clean; /pr → /commit and /ship → /pr → /commit cascades each announced every hop exactly once (the middle-link /pr stayed quiet for itself yet announced onward). Prompt-level fix judged sufficient; the deterministic PreToolUse hook stays deferred.

Not "Closes #353" — that closed on the #353 merges; this is the announcement/affirmation follow-on.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>


NorsemanSpiff:
**Addressed the review — two follow-up fixes in `ab53076`.**

**1. Announce-suppression now respects the marker's 30s lease.** The delegated-call exception in both `/commit` and `/pr` keyed on the marker's *presence*, while the intent gate keys on *freshness* (<30s). A stale marker left by a crashed delegation could suppress the announcement on a later standalone NL run while the gate correctly treated that run as standalone. Both SKILLs now require a fresh (<30s) marker to suppress; a stale one announces as normal. The exception text now points back to the gate's lease so the two checks can't drift apart again.

**2. Devjournal reconciled.** The 2026-06-26 entry's "never the announcement" line was right for the opt-out file but wrong for delegation once this PR made the marker suppress the child's self-announce. Added a 2026-07-02 entry with the final behavior and a "superseded" pointer on the older entry.

Also checked spec-portable-ai-procedures.md, CLAUDE.md, and plan-skill-nl-invocation.md — all already describe the parent-announces / child-suppresses model, so no changes there.

benjifriedman (approved):

james-baker (approved):
I read everything, looks good

Issue #492, opened 2026-07-06
e2e: version-update banner test fails once the dev server is older than 2h


james-baker:
### What happened

During a local `npm test` gate on a worktree lane, `tests/e2e/version-update.spec.ts:27` ("a feature update shows a dismissible banner") failed: the banner was visible where the test expects it held, and the file's serial mode skipped its two sibling tests. The diff under test was docs-only; the failure reproduces with no diff at all.

### Root cause

The banner's 2-hour feature hold anchors on `BUILD.builtAt` = `NEXT_PUBLIC_BUILD_TIME`, which is stamped when `next.config.ts` loads — i.e. at dev-server start. Local e2e reuses one long-running dev server (#372), so once that server is older than 2 hours the hold is already expired and the stubbed newer version surfaces the banner immediately, before the test's `page.clock.fastForward`.

The failure is intermittent only because of a masking race: `await expect(banner).toBeHidden()` resolves instantly when the banner hasn't rendered yet. A cold `/me` chunk compile delays the banner past that first poll (test passes vacuously); a warm chunk renders the banner first (test fails). Verified both ways on 2026-07-06: fails at server age >2h, 37/37 green immediately after a server restart.

CI is unaffected — `e2e.yml` runs against freshly built Vercel previews, where `builtAt` is minutes old.

### Fix direction

Pin the test's fake clock relative to the build time instead of real now, so the "held" assertion is deterministic regardless of server age — e.g. expose `builtAt` in the `GET /api/version` payload (it is already public, non-secret build metadata), then:

```ts
const { builtAt } = await (await fetch(`${baseURL}/api/version`)).json();
await page.clock.install({ time: Date.parse(builtAt) + 60_000 });
```

The existing `fastForward("02:01:00")` then crosses the hold boundary by construction. The vacuous-pass race could also be tightened while in there (assert the banner stays hidden for a beat, e.g. `fastForward` a minute before the first `toBeHidden`).

---
🤖 Filed by Claude Code

Issue #365, opened 2026-06-08
Old browsers silently fail the signup "Check code" form (macOS Yosemite / Safari ≤10)


james-baker:
### What happened

A prospective member on macOS (suspected Yosemite, Safari ≤ 10.1) opened `/signup`, entered an invite code, and clicked **Check code**. The code cleared from the input and nothing else happened — no error, no next step.

### Root cause

Next 16 / React 19 ship modern JavaScript (optional chaining `?.`, nullish coalescing `??`, etc.) that requires Safari 13.1+. Safari 10 can't parse the bundle, so React never hydrates. The submit button reverts to its native `type="submit"` behavior and does a GET to `/signup` with no `code` param, which re-renders the form with an empty field — hence "the code disappeared and nothing happened."

It works on Chrome and every current browser. Note: Sentry likely captured nothing here, because its browser SDK loads from the same bundle that failed to parse.

### Scope

This affects every interactive flow on unsupported browsers, not just signup — signup is just the first place an invited prospect hits it.

### Proposed fix

A boot-time, ES5-safe browser capability check (one that runs even when the app bundle won't parse) that surfaces a clear "please update your browser" notice instead of a silently-failing UI. The goal is to fail legibly, not to actually support these browsers.

---
🤖 Filed by Claude Code

PR #495, opened 2026-07-06
fix(myweb): old-browser-safe name halos + Baseline support policy

james-baker:
Summary: Fixes white-on-white graph name labels on browsers that ignore
paint-order on HTML text, and codifies the browser-support policy that
prevents the next such miss.

Why: A member on Chrome 109 reported an unreadable profile mini-map:
the labels' 3px canvas-colored text stroke relied on paint-order:
stroke, which Chrome honors on HTML text only since 123 (Mar 2024), so
older Chromes painted the stroke over the glyphs. The support target
was previously undocumented, so features newer than the membership's
browsers shipped by accident twice (#365 was the first).

Behavior:
- Graph name labels (profile mini-map + /myweb hover) render as a
  stroked halo copy under plain fill text — identical on current
  browsers, readable everywhere.
- New docs/strategy-browser-support.md: full support requires Baseline
  widely available; best-effort down to the hard floor (Chrome/Edge 85,
  Safari 13.1, Firefox 77 — the legacy-check blocker); CLAUDE.md
  pointers; devjournal entry.
- /about gains a member-facing "Browser support" section below The
  team, linking to web.dev/baseline (both /about external links now
  navigate in-tab).

Test Plan:
- ran `npm test` locally
- drove the dev app with Playwright: mini-map and /myweb hover labels
  render dark-on-light, halo tracks --color-canvas in dark mode, both
  layers truncate identically on a long name; /about renders the new
  section and Baseline link
- one gate run hit the pre-existing version-update 2h time-bomb,
  root-caused and filed as #492 (green on a fresh dev server)

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01VJdC6NypNJC2tUKjtec76S

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -54,6 +54,7 @@ Requires Docker Desktop. `npm run dev` auto-starts a local Supabase stack (Postg
 - Schema/API migrations use the **expand-contract pattern** (see `docs/strategy-committing.md`)
 - All pushes trigger Vercel deployment; docs-only changes (`docs/**`, root `CLAUDE.md`) skip automatically (see `docs/strategy-committing.md`)
 - Add `docs/devjournal.md` entries for decisions teammates should know about
+- All first-party code must stick to **Baseline widely available** (`docs/strategy-browser-support.md`)
 
 ## CI/CD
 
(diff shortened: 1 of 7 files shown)


james-baker:
Two commits added since the review comment:

- `c9a6ce2` feat(about): changelog entries for browser support + recent features — addresses the claude-review finding; also backfills entries for #462 (relating-dialog photos) and the program-listing improvements, which had shipped without one.
- `952a029` feat(about): tighten the browser-support and program changelog entries — James's wording pass on the two entries.

